PR #17, opened 2023-06-29
SQL schema: Move `strategy` column to front of retention policy table

amotl:
### About

By moving the `strategy` column to the front of the retention policy table, a retention policy record almost reads like natural language.

```sql
-- A policy using the REALLOCATE strategy.
INSERT INTO "ext"."retention_policy"
  (strategy, table_schema, table_name, partition_column, retention_period, reallocation_attribute_name, reallocation_attribute_value)
VALUES
  ('reallocate', 'doc', 'raw_metrics', 'ts_day', 60, 'storage', 'cold');
```

### Question

@hammerhead: Would that statement be correct, to describe this retention policy in human language? Can you correct it, if I got it wrong?

> **Reallocate** data from the `"doc"."raw_metrics"` table to machines running nodes tagged with `node.attr.storage=cold`, on partitions defined by column `ts_day`, when it is older than **60** days.

See also https://github.com/crate-workbench/cratedb-retention/pull/14#discussion_r1246068805.

### DIFF
--- a/README.md
+++ b/README.md
@@ -256,9 +256,9 @@ Add a retention policy rule using SQL.
 # A policy using the DELETE strategy.
 crash --hosts "${CRATEDB_HOST}" <<SQL
     INSERT INTO "ext"."retention_policy"
-      (table_schema, table_name, partition_column, retention_period, strategy)
+      (strategy, table_schema, table_name, partition_column, retention_period)
     VALUES
-      ('doc', 'raw_metrics', 'ts_day', 1, 'delete');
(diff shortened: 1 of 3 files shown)


hammerhead:
> **Reallocate** data from the `"doc"."raw_metrics"` table to machines running nodes tagged with `node.attr.storage=cold`, on partitions defined by column `ts_day`, when it is older than **60** days.

Yes, that is correct (instead of "machines running nodes" I would simply say "nodes", as machines could theoretically run multiple nodes with different `node.attr.storage` values).

hammerhead (approved):

PR #19, opened 2023-06-29
Add "tags" to data model, and `--tags` command-line option

amotl:
### About

By adding the concept of "tags" to the data model, it becomes capable of grouping retention policy items by different means. This effectively unlocks managing a large number of policy items, for different purposes. One example are multi-tenant operations.

### Details

While adding tags to the data model, a few spots had to be refactored and modernized. Most prominently, because more and more parts of the SQL statements are assembled by code, the file-based SQL templates have been dissolved.

- More OO: Add `RetentionPolicyStore`, `DatabaseAdapter`, and `DatabaseTagHelper`
- Dissolve file-based SQL templating


cratedb_retention/util/database.py
```diff
@@ -1,14 +1,100 @@
...
+        # Compute SQL WHERE clause for filtering by tags.
+        sql = f"SELECT * FROM {self.tablename} WHERE {where_clause};"  # noqa: S608
+        try:
+            self.database.run_sql(sql)
+        except ProgrammingError as ex:
+            if "ColumnUnknownException" in str(ex):
+                return False
+            else:
+                raise
+        return True
```
amotl, line 93: @hlcianfagna: This code helps implementing a tag system based on CrateDB's `OBJECT(IGNORED)` type. Thanks for your support on this matter, and let me know if you think it could be implemented differently/better.

cratedb_retention/util/database.py
```diff
@@ -1,14 +1,100 @@
...
+                item = f"{field_prefix}tags['{tag}'] IS NOT NULL"
+                items.append(item)
+        return items
+
+    def tags_exist(self, tags: t.List[str]):
+        """
+        Check if given tags exist in the database.
+
+        When not, no query can yield results, so we do not need to bother about
+        failing JOIN operations because if non-existing tags.
```
amotl, line 81: @hlcianfagna: I have not been able to put that into a reproducer, in order to carry the issue forward to the crate/crate issue tracker, to outline that `set error_on_unknown_object_key=false;` does not help when accessing non-existent object keys within an SQL statement which uses JOIN operations. Let me know if I can support you on this matter.

hammerhead (approved):

PR #20, opened 2023-07-02
More APIs and SQLAlchemy, less SQL

amotl:
### About

It became clear the program needed a better data model, and CLI interfaces, than only running SQL statements manually for inserting and editing retention policy rules.

### Details
- Data model API using SQLAlchemy
- CLI API using Click

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,15 +73,15 @@ More strategies can be added, and corresponding contributions are welcome.
 
 ### DELETE
 
-A basic retention policy algorithm that drops records from expired partitions.
+A retention policy algorithm that deletes records from expired partitions.
 
-```sql
--- A policy using the DELETE strategy.
-INSERT INTO "ext"."retention_policy"
-  (strategy, table_schema, table_name, partition_column, retention_period)
-VALUES
-  ('delete', 'doc', 'raw_metrics', 'ts_day', 1);
+```shell
+cratedb-retention create-policy --strategy=delete \
+  --table-schema=doc --table-name=raw_metrics \
+  --partition-column=ts_day --retention-period=1 \
+  "${CRATEDB_URI}"
 ```
+
 
 This retention policy implements the following directive.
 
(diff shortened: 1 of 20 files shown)


README.md
```diff
@@ -219,10 +245,13 @@ SET error_on_unknown_object_key=false;
 SELECT * FROM "ext"."retention_policy" WHERE tags['foo'] IS NOT NULL;
 ```
 
-Delete all retention policy records tagged with `foo`.
+Delete all retention policy records tagged with `foo` and `bar`.
 ```sql
 DELETE FROM "ext"."retention_policy" WHERE tags['foo'] IS NOT NULL;
```
hammerhead, line 250: Should be `WHERE tags['foo'] IS NOT NULL AND tags['bar'] IS NOT NULL`?

amotl: In this case, the SQL statement is `DELETE FROM "ext"."retention_policy" WHERE tags['foo'] IS NOT NULL;`, so it intends to only delete records tagged with `foo`. That it is essentially the same record which is also tagged with `bar` is just a coincidence.

amotl: Ah, you are suggesting to amend the SQL statement. Now, I am getting it. Yeah, sure, why not.

amotl: Fixed. Thank you.

cratedb_retention/setup/schema.sql
```diff
@@ -26,8 +28,7 @@ CREATE TABLE IF NOT EXISTS {policy_table.fullname} (
     "reallocation_attribute_value" TEXT,        -- Value of the node-specific custom attribute.
 
     -- Targeting a repository.
-    "target_repository_name" TEXT,              -- The name of a repository created with `CREATE REPOSITORY ...`.
+    "target_repository_name" TEXT               -- The name of a repository created with `CREATE REPOSITORY ...`.
 
-    PRIMARY KEY ("strategy", "table_schema", "table_name")
```
hammerhead, line 31: The idea behind that primary key was to prevent duplicate strategies on the same table. Too bad we don't have `UNIQUE` constraints in CrateDB.
Is there elsewhere in the code a check to prevent duplicates (i.e., for the same table, one entry with `DELETE` and 3 days retention, and another with `DELETE` and 5 days retention on the same table)?

amotl: Thanks for spotting. I haven't had the chance to outline that regression yet, but I've now created GH-26 to track it.

hammerhead (approved):

Issue #26, opened 2023-07-03
The idea behind that primary key was to prevent duplicate strategies on the same table. Too bad we don't have `UNIQUE` constraints in CrateDB.


amotl:
The idea behind that primary key was to prevent duplicate strategies on the same table. Too bad we don't have `UNIQUE` constraints in CrateDB.
Is there elsewhere in the code a check to prevent duplicates (i.e., for the same table, one entry with `DELETE` and 3 days retention, and another with `DELETE` and 5 days retention on the same table)?

_Originally posted by @hammerhead in https://github.com/crate-workbench/cratedb-retention/pull/20#discussion_r1250850198_

amotl:
Hi Niklas,

thank you for picking that up from https://github.com/crate-workbench/cratedb-retention/pull/20#pullrequestreview-1511141022. I agree with you, it is sad that we can not hold up that UNIQUEness constraint because adding a sane record editing API just needs a dedicated primary key instead.

The constraint will need to be re-added by implementing it through code, right. I will add it back on behalf of a subsequent patch.

With kind regards,
Andreas.

amotl:
GH-28 intends to resolve this issue.

amotl:
Hi again,

@hammerhead added a few valuable thoughts to this topic at https://github.com/crate-workbench/cratedb-retention/pull/28#discussion_r1257750167. Thank you very much.

> It can make sense to have multiple policies for one table. Especially, to combine the REALLOCATE strategy with others. Example: For table `x`, I want to reallocate partitions after 30 days from a hot to a warm node, and after 60 days I want to permanently delete it (or snapshot with deletion).
>
> But indeed it doesn't make sense to have a combination of the DELETE + SNAPSHOT strategies on the same table, as both will erase the data.
>
> A REALLOCATE strategy with a shorter retention period than a DELETE/SNAPSHOT strategy makes sense, though.

With kind regards,
Andreas.

PR #28, opened 2023-07-08
Add synthetic UNIQUE constraint to retention policy table

amotl:
## About

- A quick stab at GH-26.

### DIFF
--- a/cratedb_retention/store.py
+++ b/cratedb_retention/store.py
@@ -28,6 +28,10 @@ class SQLAlchemyTagHelperMixin:
     def execute(self, statement):
         pass
 
+    @abc.abstractmethod
+    def synchronize(self):
+        pass
+
     def get_tags_constraints(self, tags: t.Union[t.List[str], t.Set[str]], table_alias: t.Optional[str] = None):
         """
         Return list of SQL WHERE constraint clauses from given tags.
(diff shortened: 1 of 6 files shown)


cratedb_retention/store.py
```diff
@@ -198,3 +208,22 @@ def query(self, statement):
...
+
+    def exists(self, policy: RetentionPolicy):
+        """
+        Check if retention policy for specific table already exists.
+        """
+        table = self.table
+        selectable = sa.select(table).where(
+            table.c.table_schema == policy.table_schema,
+            table.c.table_name == policy.table_name,
+        )
```
hammerhead, line 220: It can make sense to have multiple policies for one table. Especially to combine the reallocate strategy with others.
Example: For table `x`, I want to reallocate partitions after 30 days from a hot to a cold node, and after 60 days I want to permanently delete it (or snapshot with deletion).

But indeed it doesn't make sense to have a combination of the snapshot + delete strategies on the same table, as both will erase the data.

A reallocate strategy with a shorter retention period than a delete/snapshot strategy makes sense, though.

amotl: Hi Niklas.

I was having roughly about the same thoughts, so I am happy you are referring to a specific example. Maybe we can track the new insights at GH-26, in order to come up with a better plan how to handle the constraints more specifically?

WIth kind regards,
Andreas.

hammerhead: Sounds good, let's save it for another iteration 👍

hammerhead (approved):

PR #29, opened 2023-07-08
Core: Improve assembling retention task list

amotl:
By discretely resolving the retention task list, i.e. the list of partitions to be retired, and iterating them at runtime instead of using JOINs, the program can discriminate between many more error conditions, and inform the user correspondingly. This significantly improves usability.

Currently, the error conditions are:

- Strategy not given.
- Given tag(s) do not exist.
- No matching retention policies found.
- Data table not found.
- No data to be retired.

At the same time, the refactoring simplifies the implementations of the specific retention policy task executors (delete, reallocate, snapshot).


cratedb_retention/strategy/reallocate.py
```diff
@@ -39,28 +38,30 @@ def to_sql(self):
...
-{policy_dql}
+SELECT *
+FROM partitions p
 JOIN partition_allocations a ON a.table_schema = p.table_schema
   AND a.table_name = p.table_name
   AND p.partition_ident = a.partition_ident
-  AND attributes[r.reallocation_attribute_name] <> r.reallocation_attribute_value
-WHERE {where_clause}
+  AND attributes['{reallocation_attribute_name}'] <> '{reallocation_attribute_value}'
 ORDER BY 5 ASC;
```
amotl, line 66: Dear @hammerhead,

I've spotted this `ORDER BY 5 ASC` clause within the SQL statements driving the "REALLOCATE" strategy. Can you clarify what it means, in order to improve the inline documentation? Thanks!

With kind regards,
Andreas.

hammerhead: In earlier CrateDB versions, it wasn't possible to detect (using system tables) on what type of node a partition was stored. Therefore, we needed to store the latest partition value that has been reallocated in a separate tracking table. In case multiple partitions were reallocated during one policy execution, and the reallocation of a particular partition failed, we needed to ensure we process them in chronological order to update the latest reallocated partition value correctly.

This constraint doesn't exist anymore since node attributes have been added to `sys.nodes`. However, I think it still makes sense to process partitions in chronological order to not produce (temporary) gaps. Imagine you set up a retention policy for the first time, about to delete the months of January, February, and March. If the table is queried by users at the same time. If to-be-dropped partitons aren't iterated on in a particular order, the algorithm could first drop February, then March, and then January, which would produce inconsistent gaps in the data if it was queried at the same time. Arguably, that scenario doesn't occur often and is more of an edge case.

amotl: Hi. Thank you for the swift reply. Oh, right. It is actually `ORDER BY 5 ASC`. I misinterpreted it as some kind of `LIMIT 5` statement. Now that I am recognizing that I was blatantly wrong, I feel a bit silly.

So, do you recommend any further actions here? Either/or/and:

a) Add your insights as inline comment at the corresponding spot in the code?
b) Do not use the magic number `5` to refer to some column, but refer to it by name?

hammerhead: Option b) sounds good, if easily doable. The column in question (that used to be at position `5`) is this one:

https://github.com/crate-workbench/cratedb-retention/blob/44857bf60e361a0d9cd35757f1f6771e2f3856ca/cratedb_retention/strategy/policy.sql#L16

So an  `ORDER BY TRY_CAST(p.values[r.partition_column] AS BIGINT) ASC` is what we are looking for.

amotl: Dear Niklas.

Thank you very much for sharing your expertise on this matter. I've just added 0cd896c3508a8, which slaps that `ORDER BY` on the column you referenced, which has now been named `partition_value` while refactoring the corresponding SQL statement into the application's core module.

Please note that it will now be applied on all strategies based on partitioned tables [^1]. I don't see any harm, it will resolve the edge cases you've mentioned, and it passes the test suite, so I consider this to be a good thing.

With kind regards,
Andreas.

[^1]: We actually do not have any others, up until now ;].

hammerhead (approved):

Issue #69, opened 2023-11-01
SQLAlchemy: Improve uniqueness polyfill to accept multiple columns (composite unique constraint / unique composite key)


amotl:
## About

GH-59 added a few utilities useful for additonal SQLAlchemy support, for example, a polyfill mechanism to emulate unique constraints.

## Problem

The implementation can only handle unique constraints on single columns. In practice, we need more.

## References

- https://github.com/crate-workbench/mlflow-cratedb/pull/52#pullrequestreview-1708415310.

amotl:
It has been implemented with GH-70.

PR #70, opened 2023-11-01
SQLAlchemy: Emulate composite unique constraints

amotl:
## About
On the SQLAlchemy support subsystem, improve the UNIQUE constraints polyfill to accept multiple column names, for emulating **unique composite keys**.

## References
- https://github.com/crate-workbench/cratedb-toolkit/issues/69
- https://github.com/crate-workbench/mlflow-cratedb/pull/52#pullrequestreview-1708415310

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,6 +8,9 @@
 
 - Add support for Python 3.12
 
+- SQLAlchemy: Improve UNIQUE constraints polyfill to accept multiple
+  column names, for emulating unique composite keys.
+
 
 ## 2023/10/10 0.0.1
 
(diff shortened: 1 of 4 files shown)


matriv:
Does this issues a select query on every insert/update to check that the uniqueness is respected?

amotl:
> Does this issues a select query on every insert/update to check that the uniqueness is respected?

There is no other chance to do it differently, right?

matriv:
> > Does this issues a select query on every insert/update to check that the uniqueness is respected?
> 
> There is no other chance to do it differently, right?

No, but I'm wondering if we should document it as it will impact performance.

amotl:
> I'm wondering if we should document it as it will impact performance.

Sure, it is just a polyfill for making a few humble applications "compatible". We don't actively advertise this as a feature, it is just code sitting in a repository to be reused where applicable.

matriv (approved):

PR #71, opened 2023-11-01
Add support for Python 3.12

amotl:
## About

Add Python 3.12 to the list of supported Python releases.

## References

- GH-63

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -15,7 +15,7 @@ jobs:
       fail-fast: false
       matrix:
         os: ["ubuntu-latest"]
-        python-version: ["3.11"]
+        python-version: ["3.12"]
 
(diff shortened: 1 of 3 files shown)


.github/workflows/main.yml
```diff
@@ -15,7 +15,7 @@ jobs:
       fail-fast: false
       matrix:
         os: ["ubuntu-latest"]
-        python-version: ["3.11"]
+        python-version: ["3.12"]
```
matriv, line 18: Since we declare supporting python >= 3.8 should we add all the versions in the matrix?

amotl: I think using the lower and the upper bound will be good enough. See GH-72.

matriv (approved):

PR #74, opened 2023-11-07
Add SQL runner utility `.io.sql.run_sql`

amotl:
## About

Classical yak shaving.

## Details

Similar to https://github.com/pyveci/pueblo/pull/12, this patch adds a generalized testing helper function to avoid code duplication within the educational repository [cratedb-examples](https://github.com/crate/cratedb-examples), originally conceived at [^1]. We need this utility function at many more spots, so we needed to find a good place to store it.

## Synopsis
```python
from pprint import pprint
from cratedb_toolkit.io.sql import run_sql

result = run_sql(
  dburi="crate://crate@localhost:4200",
  sql="SELECT * FROM sys.summits LIMIT 3;",
  records=True,
)
pprint(result)
```

## Notes

Note that details on the interface may change while we go, specifically on the "naming things" / DWIM side of things. If you have any suggestions, feel free to add your voice.

[^1]: https://github.com/crate/cratedb-examples/blob/main/topic/machine-learning/llm-langchain/test.py

/cc @karynzv, @marijaselakovic, @hlcianfagna, @hammerhead, @WalBeh, @andnig, @ckurze, @vvulf

### DIFF
--- a/cratedb_toolkit/util/database.py
+++ b/cratedb_toolkit/util/database.py
@@ -1,6 +1,11 @@
 # Copyright (c) 2023, Crate.io Inc.
 # Distributed under the terms of the AGPLv3 license, see LICENSE.
+import io
+import typing as t
+from pathlib import Path
+
 import sqlalchemy as sa
+import sqlparse
 from sqlalchemy.exc import ProgrammingError
 from sqlalchemy.sql.elements import AsBoolean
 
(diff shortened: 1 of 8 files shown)


matriv (approved):

cratedb_toolkit/util/database.py
```diff
@@ -22,12 +27,23 @@ def __init__(self, dburi: str):
...
+
+        sql_effective: str
+        if isinstance(sql, str):
+            sql_effective = sql
+        elif isinstance(sql, Path):
+            sql_effective = sql.read_text()
+        elif isinstance(sql, io.IOBase):
+            sql_effective = sql.read()
+        else:
+            raise TypeError("SQL statement type must be either string, Path, or IO handle")
```
seut, line 43: Just curious: couldn't this be done by using function overloads with `@typing.overload()` instead?

amotl: Hi. I have to admit I never used it. If you think the code can be improved, I will be happy to learn about it.

PR #76, opened 2023-11-10
MongoDB: Add code from `migr8` program

amotl:
## About

The `migr8` program has been maintained over at [^1]. This patch adds it to cratedb-toolkit instead, in order to streamline maintenance in terms of more regular testing and shipping.

## Details

Documentation is included. Installation works like:
```shell
pip install 'cratedb-toolkit[io]'
migr8 --help
```

A different CLI/library interface will be added on behalf of a subsequent iteration.

/cc @karynzv

[^1]: https://github.com/crate/mongodb-cratedb-migration-tool

### DIFF
--- a/.github/workflows/mongodb.yml
+++ b/.github/workflows/mongodb.yml
@@ -1,10 +1,19 @@
 ---
-name: "Tests: Common"
+name: "Tests: MongoDB"
 
 on:
-  pull_request: ~
+  pull_request:
+    branches: ~
+    paths:
+    - '.github/workflows/mongodb.yml'
+    - 'cratedb_toolkit/io/mongodb/**'
+    - 'pyproject.toml'
   push:
     branches: [ main ]
+    paths:
+    - '.github/workflows/mongodb.yml'
+    - 'cratedb_toolkit/io/mongodb/**'
+    - 'pyproject.toml'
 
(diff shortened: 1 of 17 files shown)


.github/workflows/mongodb.yml
```diff
@@ -0,0 +1,84 @@
...
+        os: ["ubuntu-latest"]
+        python-version: ["3.8", "3.12"]
+        mongodb-version: ["2", "3", "4", "5", "6", "7"]
+
+    env:
+      OS: ${{ matrix.os }}
+      PYTHON: ${{ matrix.python-version }}
+      MONGODB_VERSION: ${{ matrix.mongodb-version }}
+      # Do not tear down Testcontainers
+      TC_KEEPALIVE: true
```
amotl, line 47: It looks like `MONGODB_VERSION` may not be picked up by the test suite here. By default, it is `mongo:latest` [^1].

[^1]: https://github.com/testcontainers/testcontainers-python/blob/3add8cb09216/mongodb/testcontainers/mongodb/__init__.py#L51

amotl: `MongoDbContainerWithKeepalive` has been improved correspondingly, and split into a seperate commit, see bf156ef52b1.

PR #78, opened 2023-11-12
Testing: Generalize `KeepaliveContainer` from `CrateDBContainer`

amotl:
## About

Add generalized `KeepaliveContainer` for efficient use of Testcontainers. It will give all containers the ability to honor `TC_KEEPALIVE=true`, i.e. not shutting down the container after running the test suite. This significantly improves cycle speed.

## Details

This generalization to the Testcontainers infrastructure is needed to bring in two subsequent patches, because both will rely on it.

- GH-76
- GH-77

Because both will use this very code to run their test cases, I will not bother to add dedicated test cases for the infrastructure code.

### DIFF
--- a/cratedb_toolkit/testing/testcontainers/cratedb.py
+++ b/cratedb_toolkit/testing/testcontainers/cratedb.py
@@ -55,8 +55,7 @@ class CrateDBContainer(KeepaliveContainer, DbContainer):
     # TODO: Dual-port use with 4200+5432.
     def __init__(
         self,
-        # TODO: Use `crate/crate:nightly` by default?
-        image: str = "crate:latest",
+        image: str = "crate/crate:nightly",
         port: int = 4200,
         user: Optional[str] = None,
         password: Optional[str] = None,
--- a/cratedb_toolkit/testing/testcontainers/cratedb.py
+++ b/cratedb_toolkit/testing/testcontainers/cratedb.py
@@ -18,10 +18,12 @@ from testcontainers.core.config import MAX_TRIES
 from testcontainers.core.generic import DbContainer
 from testcontainers.core.waiting_utils import wait_for_logs
 
+from cratedb_toolkit.testing.testcontainers.util import KeepaliveContainer, asbool
+
 logger = logging.getLogger(__name__)
 
 
-class CrateDBContainer(DbContainer):
+class CrateDBContainer(KeepaliveContainer, DbContainer):
     """
     CrateDB database container.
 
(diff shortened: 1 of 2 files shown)


cratedb_toolkit/testing/testcontainers/cratedb.py
```diff
@@ -48,21 +50,21 @@ class CrateDBContainer(DbContainer):
...
     CRATEDB_PASSWORD = os.environ.get("CRATEDB_PASSWORD", "")
     CRATEDB_DB = os.environ.get("CRATEDB_DB", "doc")
-    CRATEDB_KEEPALIVE = os.environ.get("CRATEDB_KEEPALIVE", os.environ.get("TC_KEEPALIVE", False))
+    KEEPALIVE = asbool(os.environ.get("CRATEDB_KEEPALIVE", os.environ.get("TC_KEEPALIVE", False)))
 
     # TODO: Dual-port use with 4200+5432.
     def __init__(
         self,
+        # TODO: Use `crate/crate:nightly` by default?
         image: str = "crate:latest",
```
amotl, line 59: I think we should do it, to always use the most recent code for testing all downstream applications, in order to be able to discover regressions early.

cratedb_toolkit/testing/testcontainers/cratedb.py
```diff
@@ -48,21 +50,21 @@ class CrateDBContainer(DbContainer):
     CRATEDB_USER = os.environ.get("CRATEDB_USER", "crate")
     CRATEDB_PASSWORD = os.environ.get("CRATEDB_PASSWORD", "")
     CRATEDB_DB = os.environ.get("CRATEDB_DB", "doc")
-    CRATEDB_KEEPALIVE = os.environ.get("CRATEDB_KEEPALIVE", os.environ.get("TC_KEEPALIVE", False))
+    KEEPALIVE = asbool(os.environ.get("CRATEDB_KEEPALIVE", os.environ.get("TC_KEEPALIVE", False)))
 
     # TODO: Dual-port use with 4200+5432.
```
amotl, line 55: We can use this code in `def _configure()`, as exercised at GH-77. However, I don't know if that improves the situation appropriately.
```python
self.with_exposed_ports(self.port_to_expose, 5432)
```

cratedb_toolkit/testing/testcontainers/cratedb.py
amotl: Implemented with 1c49f92f8d.

Issue #58, opened 2023-10-10
Testing: Adapt "Testcontainers" implemention to `unittest`


amotl:
## About

Over here, we reported about the state of the "Testcontainers for Python" implementation, for supporting application testing with CrateDB.

- GH-57

At the issue referenced above, we will need to resolve this backlog item, in order to make the test layer usable for applications/libraries which are using Python's `unittest` module for testing.

> While a pytest-based wrapper adapter around the "Testcontainers" implementation is nice, the projects [crate-python](https://github.com/crate/crate-python) and [crash](https://github.com/crate/crash) are using Python's builtin `unittest` module. Can we also grow a `unittest`-based wrapper adapter, to be reusable by both downstream projects?

amotl:
@pilosus contributed this the other day, thank you.

- GH-82

PR #82, opened 2023-11-21
Adapt testcontainers to be testing framework agnostic

pilosus:
## About

Testing: Adapt "Testcontainers" implementation to `unittest`

## References

- GH-58
- https://github.com/crate/crash/pull/408

## Checklist

 - [x] [CLA](https://crate.io/community/contribute/cla/) is signed

### DIFF
--- a/cratedb_toolkit/testing/testcontainers/cratedb.py
+++ b/cratedb_toolkit/testing/testcontainers/cratedb.py
@@ -91,7 +91,7 @@ class CrateDBContainer(KeepaliveContainer, DbContainer):
 
         self._name = "testcontainers-cratedb"
 
-        cmd_opts = cmd_opts if cmd_opts else {}
+        cmd_opts = cmd_opts or {}
         self._command = self._build_cmd({**self.CMD_OPTS, **cmd_opts})
 
         self.CRATEDB_USER = user or self.CRATEDB_USER
@@ -110,7 +110,7 @@ class CrateDBContainer(KeepaliveContainer, DbContainer):
         for key, val in opts.items():
             if isinstance(val, bool):
                 val = str(val).lower()
-            cmd.append("-C{}={}".format(key, val))
+            cmd.append(f"-C{key}={val}")
         return " ".join(cmd)
 
     def _configure_ports(self) -> None:
--- a/cratedb_toolkit/testing/testcontainers/cratedb.py
+++ b/cratedb_toolkit/testing/testcontainers/cratedb.py
@@ -151,7 +151,6 @@ class CrateDBFixture:
         """
         Start testcontainer, used for tests set up
         """
-        logger.debug("Starting container % with args %", self.image, **kwargs)
         self.cratedb = CrateDBContainer(image=self.image, **kwargs)
         self.cratedb.start()
         self.database = DatabaseAdapter(dburi=self.get_connection_url())
--- a/cratedb_toolkit/testing/testcontainers/cratedb.py
+++ b/cratedb_toolkit/testing/testcontainers/cratedb.py
@@ -153,14 +153,14 @@ class CrateDBFixture:
     CrateDB Toolkit's `DatabaseAdapter`, agnostic of the test framework.
     """
 
-    def __init__(self, crate_version: str = "nightly"):
+    def __init__(self, crate_version: str = "nightly", **kwargs):
         self.cratedb: Optional[CrateDBContainer] = None
         self.image: str = "crate/crate:{}".format(crate_version)
         self.database: Optional[DatabaseAdapter] = None
-        self.setup()
+        self.setup(**kwargs)
 
-    def setup(self):
-        self.cratedb = CrateDBContainer(image=self.image)
+    def setup(self, **kwargs):
+        self.cratedb = CrateDBContainer(image=self.image, **kwargs)
         self.cratedb.start()
         self.database = DatabaseAdapter(dburi=self.get_connection_url())
 
@@ -168,7 +168,7 @@ class CrateDBFixture:
         if self.cratedb:
             self.cratedb.stop()
 
-    def reset(self, tables: Optional[str] = None):
+    def reset(self, tables: Optional[list] = TestDrive.RESET_TABLES):
         if tables and self.database:
             for reset_table in tables:
                 self.database.connection.exec_driver_sql(f"DROP TABLE IF EXISTS {reset_table};")
--- a/cratedb_toolkit/testing/testcontainers/cratedb.py
+++ b/cratedb_toolkit/testing/testcontainers/cratedb.py
@@ -98,9 +98,18 @@ class CrateDBContainer(KeepaliveContainer, DbContainer):
             cmd.append("-C{}={}".format(key, val))
         return " ".join(cmd)
 
-    def _configure(self) -> None:
+    def _configure_ports(self) -> None:
+        """
+        Bind all the ports exposed inside the container to the same port on the host
+        """
         ports = [*[self.port_to_expose], *self.extra_ports]
-        self.with_exposed_ports(*ports)
+        for port in ports:
+            # If random port is needed on the host, use host=None
+            # or invoke self.with_exposed_ports
+            self.with_bind_ports(container=port, host=port)
+
+    def _configure(self) -> None:
+        self._configure_ports()
         self.with_env("CRATEDB_USER", self.CRATEDB_USER)
         self.with_env("CRATEDB_PASSWORD", self.CRATEDB_PASSWORD)
         self.with_env("CRATEDB_DB", self.CRATEDB_DB)
--- a/cratedb_toolkit/testing/testcontainers/cratedb.py
+++ b/cratedb_toolkit/testing/testcontainers/cratedb.py
@@ -70,7 +70,6 @@ class CrateDBContainer(KeepaliveContainer, DbContainer):
         user: Optional[str] = None,
         password: Optional[str] = None,
         dbname: Optional[str] = None,
-        dialect: str = "crate",
         cmd_opts: Optional[dict] = None,
         **kwargs,
     ) -> None:
@@ -79,13 +78,12 @@ class CrateDBContainer(KeepaliveContainer, DbContainer):
         :param ports: optional dict that maps a port inside the container to a port on the host machine;
                       `None` as a map value generates a random port;
                       Dicts are ordered. By convention the first key-val pair is designated to the HTTP interface.
-                      Example: {4200: None, 5432: 15431} - port 4200 inside the container will be mapped
+                      Example: {4200: None, 5432: 15432} - port 4200 inside the container will be mapped
                       to a random port on the host, internal port 5432 for PSQL interface will be mapped
                       to the 15432 port on the host.
-        :param user:  optional username to access the DB; if None, try respective environment variable
-        :param password: optional password to access the DB; if None, try respective environment variable
-        :param dbname: optional database name to access the DB; if None, try respective environment variable
-        :param dialect: a string with the dialect to generate a DB URI
+        :param user:  optional username to access the DB; if None, try `CRATEDB_USER` environment variable
+        :param password: optional password to access the DB; if None, try `CRATEDB_PASSWORD` environment variable
+        :param dbname: optional database name to access the DB; if None, try `CRATEDB_DB` environment variable
         :param cmd_opts: an optional dict with CLI arguments to be passed to the DB entrypoint inside the container
         :param kwargs: misc keyword arguments
         """
@@ -102,7 +100,6 @@ class CrateDBContainer(KeepaliveContainer, DbContainer):
 
         self.port_mapping = ports if ports else {4200: None}
         self.port_to_expose, _ = list(self.port_mapping.items())[0]
-        self.dialect = dialect
 
     @staticmethod
     def _build_cmd(opts: dict) -> str:
@@ -133,11 +130,18 @@ class CrateDBContainer(KeepaliveContainer, DbContainer):
         self._configure_ports()
         self._configure_credentials()
 
-    def get_connection_url(self, host=None, dialect=None) -> str:
+    def get_connection_url(self, dialect: str = "crate", host: Optional[str] = None) -> str:
+        """
+        Return a connection URL to the DB
+
+        :param host: optional string
+        :param dialect: a string with the dialect name to generate a DB URI
+        :return: string containing a connection URL to te DB
+        """
         # TODO: When using `db_name=self.CRATEDB_DB`:
         #       Connection.__init__() got an unexpected keyword argument 'database'
         return super()._create_connection_url(
-            dialect=dialect or self.dialect,
+            dialect=dialect,
             username=self.CRATEDB_USER,
             password=self.CRATEDB_PASSWORD,
             host=host,
@@ -199,3 +203,12 @@ class CrateDBTestAdapter:
         Return a URL for CrateDB's HTTP endpoint
         """
         return self.get_connection_url(dialect="http", **kwargs)
+
+    @property
+    def http_url(self):
+        """
+        Return a URL for CrateDB's HTTP endpoint.
+
+        Used to stay backward compatible with the downstream code.
+        """
+        return self.get_http_url()
--- a/cratedb_toolkit/testing/testcontainers/cratedb.py
+++ b/cratedb_toolkit/testing/testcontainers/cratedb.py
@@ -43,6 +43,7 @@ class CrateDBContainer(KeepaliveContainer, DbContainer):
             >>> import sqlalchemy
 
             >>> cratedb_container = CrateDBContainer("crate:5.2.3")
+            >>> cratedb_container.start()
             >>> with cratedb_container as cratedb:
             ...     engine = sqlalchemy.create_engine(cratedb.get_connection_url())
             ...     with engine.begin() as connection:
@@ -65,15 +66,29 @@ class CrateDBContainer(KeepaliveContainer, DbContainer):
     def __init__(
         self,
         image: str = "crate/crate:nightly",
-        port: int = 4200,
+        ports: Optional[dict] = None,
         user: Optional[str] = None,
         password: Optional[str] = None,
         dbname: Optional[str] = None,
         dialect: str = "crate",
         cmd_opts: Optional[dict] = None,
-        extra_ports: Optional[list] = None,
         **kwargs,
     ) -> None:
+        """
+        :param image: docker hub image path with optional tag
+        :param ports: optional dict that maps a port inside the container to a port on the host machine;
+                      `None` as a map value generates a random port;
+                      Dicts are ordered. By convention the first key-val pair is designated to the HTTP interface.
+                      Example: {4200: None, 5432: 15431} - port 4200 inside the container will be mapped
+                      to a random port on the host, internal port 5432 for PSQL interface will be mapped
+                      to the 15432 port on the host.
+        :param user:  optional username to access the DB; if None, try respective environment variable
+        :param password: optional password to access the DB; if None, try respective environment variable
+        :param dbname: optional database name to access the DB; if None, try respective environment variable
+        :param dialect: a string with the dialect to generate a DB URI
+        :param cmd_opts: an optional dict with CLI arguments to be passed to the DB entrypoint inside the container
+        :param kwargs: misc keyword arguments
+        """
         super().__init__(image=image, **kwargs)
 
         self._name = "testcontainers-cratedb"
@@ -85,8 +100,8 @@ class CrateDBContainer(KeepaliveContainer, DbContainer):
         self.CRATEDB_PASSWORD = password or self.CRATEDB_PASSWORD
         self.CRATEDB_DB = dbname or self.CRATEDB_DB
 
-        self.port_to_expose = port
-        self.extra_ports = extra_ports or []
+        self.port_mapping = ports if ports else {4200: None}
+        self.port_to_expose, _ = list(self.port_mapping.items())[0]
         self.dialect = dialect
 
     @staticmethod
@@ -105,17 +120,18 @@ class CrateDBContainer(KeepaliveContainer, DbContainer):
         """
         Bind all the ports exposed inside the container to the same port on the host
         """
-        ports = [*[self.port_to_expose], *self.extra_ports]
-        for port in ports:
-            # If random port is needed on the host, use host=None
-            # or invoke self.with_exposed_ports
-            self.with_bind_ports(container=port, host=port)
+        # If host_port is `None`, a random port to be generated
+        for container_port, host_port in self.port_mapping.items():
+            self.with_bind_ports(container=container_port, host=host_port)
 
-    def _configure(self) -> None:
-        self._configure_ports()
+    def _configure_credentials(self) -> None:
         self.with_env("CRATEDB_USER", self.CRATEDB_USER)
         self.with_env("CRATEDB_PASSWORD", self.CRATEDB_PASSWORD)
         self.with_env("CRATEDB_DB", self.CRATEDB_DB)
+
+    def _configure(self) -> None:
+        self._configure_ports()
+        self._configure_credentials()
 
     def get_connection_url(self, host=None, dialect=None) -> str:
         # TODO: When using `db_name=self.CRATEDB_DB`:
--- a/cratedb_toolkit/testing/testcontainers/cratedb.py
+++ b/cratedb_toolkit/testing/testcontainers/cratedb.py
@@ -10,6 +10,10 @@
 #    WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
 #    License for the specific language governing permissions and limitations
 #    under the License.
+# This is for Python 3.7 and 3.8 to support generic types
+# like `dict` instead of `typing.Dict
+from __future__ import annotations
+
 import logging
 import os
 from typing import Optional
--- a/cratedb_toolkit/testing/testcontainers/cratedb.py
+++ b/cratedb_toolkit/testing/testcontainers/cratedb.py
@@ -58,7 +58,6 @@ class CrateDBContainer(KeepaliveContainer, DbContainer):
     KEEPALIVE = asbool(os.environ.get("CRATEDB_KEEPALIVE", os.environ.get("TC_KEEPALIVE", False)))
     CMD_OPTS = {
         "discovery.type": "single-node",
-        "cluster.routing.allocation.disk.threshold_enabled": False,
         "node.attr.storage": "hot",
         "path.repo": "/tmp/snapshots",
     }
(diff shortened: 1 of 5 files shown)


amotl (commented):
Dear Vitaly,

thank you very much for your contribution, that looks excellent. Please give us some time to thoroughly review all the details.

On CI, there may be hiccups regarding to container boot timing / availability. We know it is a hassle to get hold of those issues.

```
E       docker.errors.APIError: 500 Server Error for http+docker://localhost/v1.43/containers/a519443561406c3f60e6d0b429cfd071736191f0149c9f32cc9deb9262818adc/start: Internal Server Error ("driver failed programming external connectivity on endpoint testcontainers-cratedb (68f8422f339f8d6b40efd574ecb54861825d9ccdcbba069cfc96e4a48032833a): Bind for 0.0.0.0:4200 failed: port is already allocated")
```

-- https://github.com/crate-workbench/cratedb-toolkit/actions/runs/6950127360/job/18922807956#step:6:4850

With kind regards,
Andreas.

amotl:
We know it is a hassle to get hold of those issues. Let us know if you still have the capacity to look at those details. On this matter, I think we will have to properly configure the Docker healthcheck subsystem [^1]. I think it can use something along the lines of `curl localhost:4200/_sql` or similar -- I will have to look this up in other recipes.

https://github.com/crate-workbench/cratedb-toolkit/blob/037b8bb0d31f8c3f144d1097b619857256c48b3a/cratedb_toolkit/testing/testcontainers/cratedb.py#L98-L103

[^1]: Given that it will be available. But I think I recently stumbled upon corresponding integration with Testcontainers.

cratedb_toolkit/testing/testcontainers/cratedb.py
```diff
@@ -51,8 +56,13 @@ class CrateDBContainer(KeepaliveContainer, DbContainer):
     CRATEDB_PASSWORD = os.environ.get("CRATEDB_PASSWORD", "")
     CRATEDB_DB = os.environ.get("CRATEDB_DB", "doc")
     KEEPALIVE = asbool(os.environ.get("CRATEDB_KEEPALIVE", os.environ.get("TC_KEEPALIVE", False)))
+    CMD_OPTS = {
+        "discovery.type": "single-node",
+        "cluster.routing.allocation.disk.threshold_enabled": False,
```
amotl, line 61: This line accidentally slipped in from my personal workstation configuration, because, well, disk usage is always >80% here ;]. May I humbly ask you to remove it from this patch?

cratedb_toolkit/testing/testcontainers/cratedb.py
```diff
@@ -101,3 +135,65 @@ def _connect(self):
...
+        self.setup(**kwargs)
+
+    def setup(self, **kwargs):
+        self.cratedb = CrateDBContainer(image=self.image, **kwargs)
+        self.cratedb.start()
+        self.database = DatabaseAdapter(dburi=self.get_connection_url())
+
+    def finalize(self):
+        if self.cratedb:
+            self.cratedb.stop()
```
amotl, line 182: I called this `finalize` here, but I don't know if it was a good idea. I don't think generic pytest fixtures have a convention for "naming things" on this matter, but I will be happy to lean towards one.

What do you think about `setup`/`teardown`, or `start`/`stop`?

cratedb_toolkit/testing/testcontainers/cratedb.py
```diff
@@ -101,3 +135,65 @@ def _connect(self):
...
+        # cratedb_toolkit.io.{influxdb,mongodb}
+        '"testdrive"."demo"',
+    ]
+
+
+class CrateDBFixture:
+    """
+    A little helper wrapping Testcontainer's `CrateDBContainer` and
+    CrateDB Toolkit's `DatabaseAdapter`, agnostic of the test framework.
+    """
```
amotl, line 167: I appreciate very much bringing this into the generic `cratedb_toolkit.testing` module namespace, to make it a re-usable component for other packages.

cratedb_toolkit/testing/testcontainers/cratedb.py
```diff
@@ -101,3 +135,65 @@ def _connect(self):
...
+        f'"{EXT_SCHEMA}"."retention_policy"',
+        f'"{DATA_SCHEMA}"."raw_metrics"',
+        f'"{DATA_SCHEMA}"."sensor_readings"',
+        f'"{DATA_SCHEMA}"."testdrive"',
+        f'"{DATA_SCHEMA}"."foobar"',
+        f'"{DATA_SCHEMA}"."foobar_unique_single"',
+        f'"{DATA_SCHEMA}"."foobar_unique_composite"',
+        # cratedb_toolkit.io.{influxdb,mongodb}
+        '"testdrive"."demo"',
+    ]
```
amotl, line 160: On the other hand, contrary to `CrateDBFixture`, this section is specific to the test suite for `cratedb_toolkit`, and is not meant to be shipped with `cratedb_toolkit.testing`.

Do you see a chance to decouple this and let it be configured in `tests/conftest.py`, maybe on behalf of just a bit more pytest .request / _conf / -fixture magic, but loosely coupled, so that there is no dependency path going from `cratedb_testing` to `tests/conftest.py`, and the configuration could be somehow elegantly inverted instead?

I am not sure if I am asking for too much here, or if you can follow my thoughts easily, or if the implementation would be too complicated. Please let me know.

pilosus:
I'll address other points too. Need more time to finalize.

cratedb_toolkit/testing/testcontainers/cratedb.py
pilosus: If reset tables aren’t need in testing module and only needed in conftest, I can simply move them there. But in this case shall I still try to remove explicit import of testing/testcontainers? It’s needed for CrateDBFixture. I probably don’t see the whole picture, but utils module is still explicitly imported there, so I don’t quite get why it’s different for testing module

amotl: > If reset tables aren’t need in testing module and only needed in conftest, I can simply move them there.

This specific set of tables are only meant to be reset on behalf of the cratedb-toolkit test suite, which is not being shipped as part of the package. On the other hand, `cratedb_toolkit.testing` intends to bundle generic testing helpers/utilities/fixtures.

In this spirit, you made the right choice to put `CrateDBFixture` there, but I think the table definition / test suite configuration itself, now excellently bundled into the `TestDrive` container class, should stay in `tests/conftest.py`. However, it can't be there in isolation, because it will need to be picked up by the generic `CrateDBFixture` in some way, because this one actually orchestrates the container lifecycle. Can you figure out a way to make that happen elegantly?

> But in this case shall I still try to remove explicit import of testing/testcontainers? It’s needed for CrateDBFixture.

```
# File: cratedb_toolkit/testing/testcontainers/cratedb.py
from cratedb_toolkit.testing.testcontainers.util import KeepaliveContainer, asbool
```

```
# File: tests/conftest.py
from cratedb_toolkit.testing.testcontainers.cratedb import CrateDBFixture
```

Are you referring to one of those? I think both are fine in general. It should be free for every module to use generic utilities from `cratedb_toolkit.testing`, but not the other way round, at least import-wise.

> I probably don’t see the whole picture, but utils module is still explicitly imported there, so I don’t quite get why it’s different for testing module

Are you referring to this import?

```
# File: cratedb_toolkit/testing/testcontainers/cratedb.py
from testcontainers.core.waiting_utils import wait_container_is_ready, wait_for_logs
```

I think it is also perfectly fine. You can pull in all desired utitilities _into_ `conftest.py`, but the tricky part will be to define the test suite database connectivity configuration (`TestDrive`) _there_, and let it be picked up / consumed by the generic `CrateDBFixture` to be used properly at runtime, because it can't "import" something from `tests/conftest.py`.

amotl:
>> Bind for 0.0.0.0:4200 failed: port is already allocated
>
> is probably because of the port 4200 being used in the CI this way:

Ah I see. That makes sense.

> I think I can either revert back to the random port binding, or simply use another port in the fixture (which is probably better that the former option)

I think I'd prefer the random port assignment again for now, because it seems to be the standard way Testcontainers is doing it. By keeping up that paradigm, one has to be rigid to exclusively pick up the DBURI provided by the test machinery, and "not just sloppily assume there will be something on localhost:4200 or another fixed port anyway" ;].

What I don't like about the random port assignment paradigm is that is usually a bit more complicated to re-use test containers across subsequent test runs, and so it is mostly forgotten to be implemented on behalf of corresponding test container managers.

However, in our little corner, there is the generic `TC_KEEPALIVE=true` mechanism, which will keep the container running, so there is actually no penalty for subsequent test cycles. Personally, I am keeping this container running, dedicatedly for test harness use, and I still have the freedom to whip up ad hoc instances of CrateDB on the vanilla ports 4200/5432 for other purposes. Restarting the CrateDB container on each test run is unbearable.

> I'll address other points too. Need more time to finalize.

Thank you very much. Please take your time.

cratedb_toolkit/testing/testcontainers/cratedb.py
pilosus: So, basically, I've made `CrateDBFixture.reset` method to take reset tables in as a parameter without hardcoding them anyhow. That means, `CrateDBFixture` can still live in the `cratedb_toolkit/testing/testcontainers/cratedb.py`, but the reset tables are defined in the `tests/contest.py` along with the fixtures that use `CrateDBFixture`. I hope this solves your concern with the loose coupling.

pilosus:
@amotl I've come up with the compromise solution to the hardcoded/random port binding for the `testcontainers`: the ports are to be taken in a a dict mapping inside container port to the port on the host machine. If the host port is `None`, a random host port will be generated. This follows Docker SDK for Python [conventions](https://docker-py.readthedocs.io/en/stable/containers.html) carefully. Some reasonable default when no ports dict supplied are also given (4200 to a random port). See how it works in the [crash PR](https://github.com/crate/crash/pull/408).

Also, I've fixed other points you've given above. Added a bit more docs, prettified the code.
Please, let me know if I missed something.

cratedb_toolkit/testing/testcontainers/cratedb.py
amotl: Talking about "naming things", if you have a better idea for the name, in order to better disambiguate from pytest's notion of "fixtures", because it is actually a testframework-agnostic adapter/wrapper around, well, DatabaseAdapter [^1] and CrateDBContainer.

[^1]: Also eventually to be renamed to `CrateDBClientAdapter` or something different.

amotl: Would `CrateDBTestAdapter` be a more sensible name?

tests/conftest.py
```diff
@@ -72,17 +44,18 @@ def cratedb_service():
     Provide a CrateDB service instance to the test suite.
     """
     db = CrateDBFixture()
-    db.reset()
+    db.start(ports={CRATEDB_HTTP_PORT: None}, cmd_opts=CRATEDB_SETTINGS)
+    db.reset(tables=RESET_TABLES)
     yield db
-    db.finalize()
+    db.stop()
```
amotl, line 50: Excellent. This reads much better now. Thanks.

tests/conftest.py
```diff
@@ -72,17 +44,18 @@ def cratedb_service():
...
 
 
 @pytest.fixture(scope="function")
 def cratedb(cratedb_service):
     """
     Provide a fresh canvas to each test case invocation, by resetting database content.
     """
-    cratedb_service.reset()
+    cratedb_service.reset(tables=RESET_TABLES)
```
amotl, line 58: Ahja, excellent!

cratedb_toolkit/testing/testcontainers/cratedb.py
amotl: Yeah this was exactly what I was dreaming up regarding separation of concerns. Thank you very much.

cratedb_toolkit/testing/testcontainers/cratedb.py
```diff
@@ -101,3 +150,53 @@ def _connect(self):
...
+        if self.cratedb:
+            return self.cratedb.get_connection_url(*args, **kwargs)
+        return None
+
+    @property
+    def http_url(self):
+        """
+        Return a URL for HTTP interface
+        """
+        return self.get_connection_url(dialect="http")
```
amotl, line 202: #### Re.: Naming things

`get_connection_url` and `http_url` may also be renamed. However, it is also not relevant for this patch, but merely meant to encourage you that you can easily come up with ideas for renaming, because `cratedb-toolkit` is very young, and most of its code didn't see more than two or three iterations yet, so it offers many opportunities in this regard.

amotl (approved):
Thank you again. Approved 💯. I will also add @seut and @matriv as secondary reviewers this time.

cratedb_toolkit/testing/testcontainers/cratedb.py
pilosus: Given that we have `DatabaseAdapter`, I like the `CrateDBTestAdapter` name!

cratedb_toolkit/testing/testcontainers/cratedb.py
pilosus: When naming these, we can go in two directions:
1. rely on the `testcontainers` naming conventions, e.g. if it's a connection URL that we get using `testcontainers.core.generic.DbContainer._create_connection_url` and the interface sugests to implement `get_connection_url`, then `get_connection_url` seems very natural to me.
2. rely on `CrateDB` naming. In this case we can say it's getting the [host](https://cratedb.com/docs/crate/crash/en/latest/run.html) url or the [interface](https://cratedb.com/docs/crate/reference/en/5.5/interfaces/index.html#) url. So that we can rename `get_connection_url` to `get_interface_url` or `get_host_url`

Tbh, I think the current naming is ok. If someone needs to go deeper into `testcontainers` base classes, `get_connection_url` looks better to me.

pilosus: I'd rename `http_url` though to `get_http_url` to make it consistent with the `get_connection_url`.

cratedb_toolkit/testing/testcontainers/cratedb.py
amotl: Wonderful. Let us rename it on behalf of a subsequent patch.

cratedb_toolkit/testing/testcontainers/cratedb.py
amotl: I agree with your assessment to follow the naming conventions of Testcontainers. Maybe we can have an inline docstring comment on the `get_connection_url` method to tell the readers that it is actually an SQLAlchemy database connection uri that will be returned here?

pilosus: `get_connection_url` docstring says: `Return a URL for SQLAlchemy DB engine`. Do you want it to put it differently anyhow?

amotl: Ah. No it is perfect. Apologies if I missed it.

cratedb_toolkit/testing/testcontainers/cratedb.py
amotl: I see you already renamed it properly, thanks.

I proposed to change it later, because I feared this would add too much noise to this patch because the symbol would need to be changed at too many places. But, of course, it turns out that it already has been nicely decoupled from the test cases on behalf of the `cratedb` and `cratedb_service` pytest fixtures, so my fears were unfounded.

cratedb_toolkit/testing/testcontainers/cratedb.py
```diff
@@ -51,44 +57,87 @@ class CrateDBContainer(KeepaliveContainer, DbContainer):
...
         dialect: str = "crate",
+        cmd_opts: Optional[dict] = None,
         **kwargs,
     ) -> None:
+        """
+        :param image: docker hub image path with optional tag
+        :param ports: optional dict that maps a port inside the container to a port on the host machine;
+                      `None` as a map value generates a random port;
+                      Dicts are ordered. By convention the first key-val pair is designated to the HTTP interface.
+                      Example: {4200: None, 5432: 15431} - port 4200 inside the container will be mapped
```
matriv, line 82:
```suggestion
                      Example: {4200: None, 5432: 15432} - port 4200 inside the container will be mapped
```

cratedb_toolkit/testing/testcontainers/cratedb.py
```diff
@@ -51,44 +57,87 @@ class CrateDBContainer(KeepaliveContainer, DbContainer):
...
+                      to a random port on the host, internal port 5432 for PSQL interface will be mapped
+                      to the 15432 port on the host.
+        :param user:  optional username to access the DB; if None, try respective environment variable
+        :param password: optional password to access the DB; if None, try respective environment variable
+        :param dbname: optional database name to access the DB; if None, try respective environment variable
+        :param dialect: a string with the dialect to generate a DB URI
```
matriv, line 88: why do we want to expose `dialect` as parameter? is there a use case to change it?

cratedb_toolkit/testing/testcontainers/cratedb.py
```diff
@@ -51,44 +57,87 @@ class CrateDBContainer(KeepaliveContainer, DbContainer):
...
     ) -> None:
+        """
+        :param image: docker hub image path with optional tag
+        :param ports: optional dict that maps a port inside the container to a port on the host machine;
+                      `None` as a map value generates a random port;
+                      Dicts are ordered. By convention the first key-val pair is designated to the HTTP interface.
+                      Example: {4200: None, 5432: 15431} - port 4200 inside the container will be mapped
+                      to a random port on the host, internal port 5432 for PSQL interface will be mapped
+                      to the 15432 port on the host.
+        :param user:  optional username to access the DB; if None, try respective environment variable
```
matriv, line 85: here and below, maybe replace `respective` with the actual name of the env variable.

matriv (commented):
thanks for your work @pilosus! I've left a few comments.

cratedb_toolkit/testing/testcontainers/cratedb.py
amotl: I am not 100% sure, but I think when passed `"http"` instead of `None`, the corresponding utility function will return the HTTP-based connection URL, and not the SQLAlchemy-based connection URL, in the same manner as `get_http_url()` does it.

The `dialect` argument has already been part of the implementation before the refactoring, and I think it is part of the official/designated constructor signature. See [^1].

[^1]: https://github.com/crate-workbench/cratedb-toolkit/blob/v0.0.2/cratedb_toolkit/testing/testcontainers/cratedb.py#L61

matriv: `dialect` is normally used to use a DB specific dialect instead of a more generic one, e.g. psql vs jdbc, it's what we use to set `crate` instead of `psql` for the cases that we want to use the `crate` jdbc driver instead of the psql one. `http` vs binary `psql` protocol is not a dialect, but a protocol.

matriv: Not saying that it's part of the PR, just mentioning that it seems weird to me to expose a "dialect".

cratedb_toolkit/testing/testcontainers/cratedb.py
pilosus: Fixed

cratedb_toolkit/testing/testcontainers/cratedb.py
pilosus: Fixed

cratedb_toolkit/testing/testcontainers/cratedb.py
pilosus: Actually, I've double checked the dialect thing and you are right @matriv - no need to have it in the class constructor. It's enough to have it in the `get_connection_url` as an argument with the default value `crate`. Gonna fix that

matriv (approved):
thank you @pilosus !

cratedb_toolkit/testing/testcontainers/cratedb.py
```diff
@@ -51,44 +57,91 @@ class CrateDBContainer(KeepaliveContainer, DbContainer):
     CRATEDB_PASSWORD = os.environ.get("CRATEDB_PASSWORD", "")
     CRATEDB_DB = os.environ.get("CRATEDB_DB", "doc")
     KEEPALIVE = asbool(os.environ.get("CRATEDB_KEEPALIVE", os.environ.get("TC_KEEPALIVE", False)))
+    CMD_OPTS = {
+        "discovery.type": "single-node",
+        "node.attr.storage": "hot",
```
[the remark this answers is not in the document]
amotl: Yeah, well spotted. It is indeed a configuration detail which is specific to the test cases for `cratedb_toolkit.retention`. 

@pilosus: Do you think we can improve this spot, so that corresponding configuration settings can be defined on behalf of the snippet in `conftest.py`? This time, it will probably not be so easy, because the test adapter will already need this information at startup time.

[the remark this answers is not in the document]
amotl: > This time, it will probably not be so easy, because the test adapter will already need this information at startup time. Maybe you have an idea how to handle this elegantly?
>
> While being at it: Of course, it would not just be about the specific `node.attr.storage` parameter, but about any other parameters as well.

Ah, I see there is already `cmd_opts` on the constructor. So, `{"node.attr.storage": "hot"}` would just need to be moved over there.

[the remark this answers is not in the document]
pilosus: Exactly. And given that we use [dict merge](https://github.com/crate-workbench/cratedb-toolkit/pull/82/files#diff-a269e42ec7e05ba66ed28ac184deeaa00fb81a0db3ccb47d34635c654f1b68bcR95), you can even override the default settings

```
In [1]: from cratedb_toolkit.testing.testcontainers.cratedb  import CrateDBContainer

In [2]: c = CrateDBContainer(cmd_opts={"node.attr.storage": "cold"})

In [3]: c._command
Out[3]: '-Cdiscovery.type=single-node -Cnode.attr.storage=cold -Cpath.repo=/tmp/snapshots'
```

I'll remove the `node.attr.storage` from the default though

cratedb_toolkit/testing/testcontainers/cratedb.py
```diff
@@ -51,44 +57,91 @@ class CrateDBContainer(KeepaliveContainer, DbContainer):
...
         super().__init__(image=image, **kwargs)
 
-        self._name = "testcontainers-cratedb"  # -{os.getpid()}
-        self._command = "-Cdiscovery.type=single-node -Ccluster.routing.allocation.disk.threshold_enabled=false"
-        # TODO: Generalize by obtaining more_opts from caller.
-        self._command += " -Cnode.attr.storage=hot"
-        self._command += " -Cpath.repo=/tmp/snapshots"
+        self._name = "testcontainers-cratedb"
+
+        cmd_opts = cmd_opts if cmd_opts else {}
```
amotl, line 94: Just a nit: I usually write this like that. Is it a good idea?

```suggestion
        cmd_opts = cmd_opts or {}
```

cratedb_toolkit/testing/testcontainers/cratedb.py
```diff
@@ -51,44 +57,91 @@ class CrateDBContainer(KeepaliveContainer, DbContainer):
...
+    @staticmethod
+    def _build_cmd(opts: dict) -> str:
+        """
+        Return a string with command options concatenated and optimised for ES5 use
+        """
+        cmd = []
+        for key, val in opts.items():
+            if isinstance(val, bool):
+                val = str(val).lower()
+            cmd.append("-C{}={}".format(key, val))
```
amotl, line 113: Is there a reason not to use f-strings already?
```suggestion
            cmd.append(f"-C{key}={val}")
```

amotl (commented):
Hi Vitaly. Thanks once again for your work on this. I've added two more nitpicky suggestions ;].

cratedb_toolkit/testing/testcontainers/cratedb.py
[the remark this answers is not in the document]
amotl: Thanks. What do you think about `path.repo`, @seut? It could be convenient for testing to have it configured by default.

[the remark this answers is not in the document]
pilosus: @seut @amotl hmm, we I delete `node.attr.storage=hot`
tests: 
- `tests/retention/test_cli.py::test_run_delete_basic` 
- `tests/retention/test_cli.py::test_run_delete_dryrun`
- `tests/retention/test_cli.py::test_run_reallocate`
hang.

Here are the logs:
```
2023-11-24 17:39:14,494 [cratedb_toolkit.retention.setup.schema] INFO    : Installing retention policy bookkeeping table at database 'crate://crate:REDACTED@localhost:33018', table TableAddress(schema='testdrive-ext', table='retention_policy')
2023-11-24 17:39:14,902 [cratedb_toolkit.retention.store     ] INFO    : Connecting to database crate://crate:REDACTED@localhost:33018, table "testdrive-ext"."retention_policy"
Waiting to be ready...
2023-11-24 17:39:15,366 [testcontainers.core.waiting_utils   ] INFO    : Waiting to be ready...
```

Interestingly enough, `test_run_delete_with_tags_match` works well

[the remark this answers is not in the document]
amotl: Yeah, the test cases defined for `cratedb_toolkit.retention` need this setting to be configured, so it will need to go into `tests/conftest.py` somehow. On the other hand, it should not be part of the generic configuration. That's yet another cliff we need to take.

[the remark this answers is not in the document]
pilosus: Given the whole chain of the nested fixtures with different scopes, it will either require some more time from me next week, or a simpler solution with `cratedb` override for `tests/retention/conftest.py` that have `node.attr.storage=hot` in it.

cratedb_toolkit/testing/testcontainers/cratedb.py
pilosus: Done

cratedb_toolkit/testing/testcontainers/cratedb.py
pilosus: Fixed.
Alhough I personally not a big fan of f-strings, because sometimes people abuse their flexibility with larger expressions and, thus, damage readability. But it's a personal choice rather than anything else.

cratedb_toolkit/testing/testcontainers/cratedb.py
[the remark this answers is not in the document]
amotl: We can easily handle that on a later iteration, and/or discuss possible solutions beforehand.

PR #111, opened 2024-02-07
Sandbox: Add `all` extra to dev install command

surister:
Without it some test fail when running all.

### DIFF
--- a/doc/sandbox.md
+++ b/doc/sandbox.md
@@ -18,7 +18,7 @@ source .venv/bin/activate
 
 Install project in sandbox mode.
 ```shell
-pip install --editable='.[io,test,develop]'
+pip install --editable='.[io,test,develop,cloud]'
 ```
 
 Run tests. `TC_KEEPALIVE` keeps the auxiliary service containers running, which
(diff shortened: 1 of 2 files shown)


amotl (approved):
Apparently missed to update docs correspondingly. This area is a work-in-progress, feel free to improve any time, because the project is still in its infancy. As long as we are on 0.0.x, nothing should hold the package back on breaking things, there are no promises at all.

Thanks for spotting and fixing! 💯

doc/sandbox.md
```diff
@@ -18,7 +18,7 @@ source .venv/bin/activate
 
 Install project in sandbox mode.
 ```shell
-pip install --editable='.[io,test,develop]'
+pip install --editable='.[io,test,develop,cloud]'
```
amotl, line 21: Maybe the project metadata gained an `all` target already, which could be used instead?
```suggestion
pip install --editable='.[all]'
```

amotl: Here we go.

https://github.com/crate-workbench/cratedb-toolkit/blob/6d03b425afe565aea8997833ef238ef7265182fd/pyproject.toml#L100-L102

amotl: So, I figure that instruction might be more appropriate?
```suggestion
pip install --editable='.[all,test,develop]'
```

amotl: Implemented on behalf of e2af46e08a.

PR #135, opened 2024-04-09
Dependencies: Unpin upper version bound of `dask`

amotl:
## Details

Otherwise, compatibility issues can not be resolved quickly, like with Python 3.11.9.

## References
- https://github.com/dask/dask/issues/11038
- https://github.com/pycaret/pycaret/issues/3960#issuecomment-2045881422

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,9 @@
 
 
 ## Unreleased
+- Dependencies: Unpin upper version bound of `dask`. Otherwise, compatibility
+  issues can not be resolved quickly, like with Python 3.11.9.
+  https://github.com/dask/dask/issues/11038
 
 ## 2024/03/22 v0.0.9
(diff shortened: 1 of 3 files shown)


amotl:
The package selects dask 2024.3.1, because dask-expr 1.0.5 says so.
```
pip install --upgrade dask

dask-expr 1.0.5 requires dask==2024.3.1, but you have dask 2024.4.1 which is incompatible.
```

However, selecting dask 2024.3.1 on the most recent Python 3.11 is a bad idea, as we know from https://github.com/pycaret/pycaret/issues/3960#issuecomment-2045881422.
```python
TypeError: descriptor '__call__' for 'type' objects doesn't apply to a 'property' object
```
-- https://github.com/crate-workbench/cratedb-toolkit/actions/runs/8621823379/job/23631548577?pr=135#step:5:131

PR #88, opened 2023-12-01
Add CFR and WTF diagnostics programs

amotl:
### WTF
There are many smart queries to leverage information from CrateDB's internal `sys.*` tables. This subsystem aims to bundle and collect them, in order to unlock easy access from CLI and HTTP interfaces, and to be able to use them as building blocks for other software components.

### CFR
Export and import CrateDB system tables conveniently.

## Status

It is still a work in progress, and needs more attention and love. What can be done right now is outlined within the preview section. Any kind of help to expand this is much appreciated.

## Documentation

Rendered in preview mode.

- [CrateDB CFR](https://cratedb-toolkit--88.org.readthedocs.build/cfr/)
- [CrateDB WTF](https://cratedb-toolkit--88.org.readthedocs.build/wtf/)

Both documents include directives to outline how operations work using Docker. However, the program can also be installed natively, also while not on PyPI yet. Please use one of those commands to acquire the software. When using Docker, make sure to use the `cratedb-toolkit:pr-88` image on GHCR for preview purposes.

```
# Native
pip install --upgrade 'cratedb-wtf @ git+https://github.com/crate-workbench/cratedb-toolkit@cratedb-wtf'
```

```
# OCI
docker pull ghcr.io/crate-workbench/cratedb-toolkit:pr-88
```

## Help

```
ctk cfr --help
ctk wtf --help
```

## Backlog
- [x] Documentation.
- [x] Software tests.
- [x] Cherry-pick a few fixes from GH-153.
- ~~[ ] Handovers.~~
- [x] Derive future improvements from review process.

PR #153, opened 2024-05-30
cfr: Add support to directly export to an archive file

seut:
Some CFR improvements:
 - Add support to directly export the system tables into a `tgz` file

### DIFF
--- a/cratedb_toolkit/cfr/cli.py
+++ b/cratedb_toolkit/cfr/cli.py
@@ -6,7 +6,7 @@ import sys
 import click
 from click_aliases import ClickAliasedGroup
 
-from cratedb_toolkit.cfr.systable import SystemTableExporter, SystemTableImporter
+from cratedb_toolkit.cfr.systable import Archive, SystemTableExporter, SystemTableImporter
 from cratedb_toolkit.util.cli import (
     boot_click,
     error_logger,
@@ -46,8 +46,20 @@ def cli(ctx: click.Context, cratedb_sqlalchemy_url: str, verbose: bool, debug: b
 def sys_export(ctx: click.Context, target: str):
     cratedb_sqlalchemy_url = ctx.meta["cratedb_sqlalchemy_url"]
     try:
-        stc = SystemTableExporter(dburi=cratedb_sqlalchemy_url, target=path_from_url(target))
+        target_path = path_from_url(target)
+        stc = SystemTableExporter(dburi=cratedb_sqlalchemy_url, target=target_path)
+
+        archive = None
+        if target_path.name.endswith(".tgz") or target_path.name.endswith(".tar.gz"):
+            archive = Archive(stc)
+
         path = stc.save()
+
+        if archive is not None:
+            path = archive.make_tarfile()
+            archive.close()
+            logger.info(f"Created archive file {target}")
+
         jd({"path": str(path)})
     except Exception as ex:
         error_logger(ctx)(ex)
--- a/tests/cfr/test_cli.py
+++ b/tests/cfr/test_cli.py
@@ -1,7 +1,9 @@
 import json
+import os.path
 import re
 import shutil
 import sys
+import tarfile
 
 import tests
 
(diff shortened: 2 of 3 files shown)


cratedb_toolkit/cfr/cli.py
```diff
@@ -46,8 +46,8 @@ def cli(ctx: click.Context, cratedb_sqlalchemy_url: str, verbose: bool, debug: b
 def sys_export(ctx: click.Context, target: str):
     cratedb_sqlalchemy_url = ctx.meta["cratedb_sqlalchemy_url"]
     try:
-        stc = SystemTableExporter(dburi=cratedb_sqlalchemy_url, target=path_from_url(target))
-        path = stc.save()
+        stc = SystemTableExporter(dburi=cratedb_sqlalchemy_url)
+        path = stc.save(path_from_url(target))
```
amotl, line 50: The previous interface is by design. The program should not accept a full path to the tarfile, but a target directory instead, like it is outlined in the [corresponding documentation section](https://cratedb-toolkit--88.org.readthedocs.build/cfr/#usage).

The reason for this design is because the `.save()` operation should create a filename including a timestamp. In this way, it can be easily invoked multiple times. In this spirit, it is the foundation for a new feature to be added on another iteration, implementing `ctk cfr sys-export --interval=10s file:///var/ctk/cfr`, which will yield multiple files, and as such, accepting an absolute path to a single tarfile would lack the corresponding capacity.

amotl: In fact, the routine already drops files into a timestamped **_folder_** instead:
```python
path = self.target / self.info.cluster_name / timestamp / "sys"
```

The idea was to just use that path up until `self.target / self.info.cluster_name / timestamp` and wrap its contents into a tarfile using a compound name of cluster name + timestamp, at least for single-shot operations.

For multi-shot operations, the idea was to just wrap _all subdirectories_ up to `self.target / self.info.cluster_name` into a tarfile, for better »naming things« and multi-tenant operations, co-designed together with @WalBeh. Thanks!

Maybe just use the same paradigm/schema for **both variants, where single-shot is just a special case?** This makes decoding CFR snapshots more straight-forward, because the client/decoder (`cfr import`) has an easier job, because it can always expect the same layout of the ingredients of such a tarfile, regardless whether it was a single-shot or a multi-shot CFR recording.

pyproject.toml
```diff
@@ -102,6 +102,11 @@ dependencies = [
   "tqdm<5",
   'typing-extensions<5; python_version <= "3.7"',
   "uvicorn<0.25",
+  "yarl",
+  "numpy",
+  "pyyaml",
+  "pandas",
+  "pyarrow"
```
amotl, line 109: I would have expected the linter croaks because the dependencies have not been sorted correctly. Anyway. However, please let's add corresponding version limiters.
```suggestion
  "yarl<1.10",
  "numpy<1.27",
  "pandas<3",
  "pyarrow<17",
  "pyyaml<7",
```

tests/cfr/assets/sys_operations.sql
```diff
@@ -1,4 +1,4 @@
-CREATE TABLE IF NOT EXISTS "sys-operations" (
+CREATE TABLE IF NOT EXISTS "crate"."sys_operations" (
```
amotl, line 1: The idea was that the user can specify the schema, for example like this:

```shell
export CRATEDB_SQLALCHEMY_URL=crate://localhost/?schema=foobar
```

-- https://cratedb-toolkit--88.org.readthedocs.build/cfr/#synopsis

tests/cfr/test_cli.py
```diff
@@ -43,10 +45,48 @@ def test_cfr_cli_export(cratedb, tmp_path, caplog):
     assert filenames(path) == ["data", "schema"]
 
     schema_files = filenames(path / "schema")
-    data_files = filenames(path / "schema")
+    data_files = filenames(path / "data")
```
amotl, line 48: Thank you for spotting this.

amotl (changes requested):
Thank you so much for contributing to the last mile to complete this feature in its first iteration. I've added a few suggestions, and extensive thoughts and rationales why we would like to keep the previous interface as outlined on the [CFR documentation page](https://cratedb-toolkit--88.org.readthedocs.build/cfr/).

cratedb_toolkit/cfr/cli.py
seut: > In this way, the .save() operation can be easily invoked multiple times. 

I see, didn't think of this design approach. I wondered already why the ctor should take an export path.

> Also, letting the user specify the tarfile name can create ambiguities on the receivers end, specifically if you are receiving multiple CFR recordings over the course of a longer session, and eventually from multiple users/customers.

I'm not really sure why letting the user specify the tarfile name would violate this design. I'm also not sure if this tool should really be used for record over time, also why should it be used for multiple users?
If this is really what you have in mind, we should come up with a different tool additionally. 
My main goal is to provide a one-shot way of gathering all cluster information, mainly that users can run it and send it to us while asking for any support.
I tend to feel that we are trying to solve different expectations with the same tool/code, maybe we need to split it in various parts then.

tests/cfr/assets/sys_operations.sql
seut: Yeah,  this was more tbd. I'm fine to drop this.

pyproject.toml
seut: Sure can do. Just a reminder, this is a draft not a final PR ;)

cratedb_toolkit/cfr/cli.py
amotl: Without reviewing in detail yet, introducing a dedicated `class Archive`, paired with your comment

> whatever is saved will be archived

.... sounds excellent. Thanks a stack for refactoring those concerns appropriately, so they can be used flexibly, without introducing too much of any code smells like making one kind of operation (single-shot vs. multi-shot) feel like a hack.

I think for the first iteration, you've completed the implementation of the CFR parts of GH-88 in the best way possible, absolutely like I intended it to be, just hampered because running out of steam. Thank you so much for taking over.

cratedb_toolkit/cfr/cli.py
```diff
@@ -46,8 +46,20 @@ def cli(ctx: click.Context, cratedb_sqlalchemy_url: str, verbose: bool, debug: b
...
+        archive = None
+        if target_path.name.endswith(".tgz") or target_path.name.endswith(".tar.gz"):
+            archive = Archive(stc)
+
         path = stc.save()
+
+        if archive is not None:
+            path = archive.make_tarfile()
+            archive.close()
+            logger.info(f"Created archive file {target}")
```
amotl, line 61: I like that approach very much, perfect [DWIM](https://en.wikipedia.org/wiki/DWIM).

tests/cfr/assets/sys_operations.sql
seut: I dropped this now. Should be done in a separate PR, but the effort to have a fallback schema may be a bit high. Not really sure if this is worth it.

amotl (approved):
Thanks a stack. 💯

PR #163, opened 2024-06-10
Dependencies: Migrate from `crate[sqlalchemy]` to `sqlalchemy-cratedb`

amotl:
## About
The [CrateDB SQLAlchemy dialect](https://github.com/crate-workbench/sqlalchemy-cratedb) needs more love, so it was separated from the [DBAPI HTTP driver](https://github.com/crate/crate-python). This patch verifies and concludes the migration.

## References
- https://github.com/crate/roadmap/issues/85

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -90,11 +90,11 @@ dependencies = [
   "colorama<1",
   "colorlog",
   "crash",
-  "crate[sqlalchemy]>=0.34",
+  "crate==1.0.0dev0",
   'importlib-metadata; python_version <= "3.7"',
   "python-dotenv<2",
   "python-slugify<9",
-  "sqlalchemy",
+  "sqlalchemy-cratedb",
   "sqlparse<0.6",
   'typing-extensions<5; python_version <= "3.7"',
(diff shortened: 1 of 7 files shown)


pyproject.toml
```diff
@@ -90,11 +90,11 @@ dependencies = [
   "colorama<1",
   "colorlog",
   "crash",
-  "crate[sqlalchemy]>=0.34",
+  "crate @ git+https://github.com/crate/crate-python.git@no-sqlalchemy",
```
amotl, line 93: The `crate` package will probably need to be published before being able to conclude the migration. For example, let's use/ focus on `crate[sqlalchemy]`.

amotl: Published crate-python@no-sqlalchemy [^1] as crate-1.0.0.dev0 [^2], in order to satisfy downsteam dependencies, without tripping any existing users of `crate`.

[^1]: https://github.com/crate/crate-python/tree/no-sqlalchemy
[^2]: https://pypi.org/project/crate/1.0.0.dev0/

PR #172, opened 2024-06-13
Use sqlalchemy's 'quote' function to quote table names and fix table quoting in cfr exporter

### LINKED ISSUE #168: [infra] Improve `quote_table_name` to accept full-qualified table identifiers
> @seut: Do you think this routine needs to be improved? You mentioned something about "quoting going south". Maybe the root cause is here, because the routine may only handle a few situations correctly?

> Yes the issue here is that it will not quote anything if the ident contains a `.` like `foo.bar`. This looks a bit wrong as it normally should quote an ident which contains dots, this would be a valid table name for PG but not for CrateDB which forbids using a `.` inside a table identifier (see https://cratedb.com/docs/crate/reference/en/latest/general/ddl/create-table.html#naming-restrictions).
But then I wonder why you're not using [sqlalchemy.sql.expression.quoted_name](https://docs.sqlalchemy.org/en/20/core/sqlelement.html#sqlalchemy.sql.expression.quoted_name) which works similar afaik.
That it tries to detect a FQ identifier, splits it and quotes all parts separately, would be a bit custom but needed if the schema and table idents are not quoted dedicated.

_Originally posted by @seut in https://github.com/crate-workbench/cratedb-toolkit/pull/88#discussion_r1627267605_

seut:
See commits.

Closes #168.

### DIFF
--- a/tests/util/database.py
+++ b/tests/util/database.py
@@ -1,0 +1,23 @@
+import pytest
+
+from cratedb_toolkit.util import DatabaseAdapter
+
+
+def test_quote_relation_name():
+    database = DatabaseAdapter(dburi="crate://localhost")
+    assert database.quote_relation_name("my_table") == "my_table"
+    assert database.quote_relation_name("my-table") == '"my-table"'
+    assert database.quote_relation_name("MyTable") == '"MyTable"'
+    assert database.quote_relation_name('"MyTable"') == '"MyTable"'
+    assert database.quote_relation_name("my_schema.my_table") == "my_schema.my_table"
+    assert database.quote_relation_name("my-schema.my_table") == '"my-schema".my_table'
+    assert database.quote_relation_name('"wrong-quoted-fqn.my_table"') == '"wrong-quoted-fqn.my_table"'
+    assert database.quote_relation_name('"my_schema"."my_table"') == '"my_schema"."my_table"'
+    # reserved keyword must be quoted
+    assert database.quote_relation_name("table") == '"table"'
+
+
+def test_quote_relation_name_with_invalid_fqn():
+    database = DatabaseAdapter(dburi="crate://localhost")
+    with pytest.raises(ValueError):
+        database.quote_relation_name("my-db.my-schema.my-table")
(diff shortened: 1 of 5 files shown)


seut:
Moving back to draft as there seem to be issues.

amotl:
> Moving back to draft as there seem to be issues.

It might be totally unrelated, but it also might be related to unfortunate side-effects of shortcomings within the test suite, as only three test cases are failing. Let me bring in this patch first, and then have another look after rebasing.

- https://github.com/crate-workbench/sqlalchemy-cratedb/pull/10

seut:
> It might be totally unrelated, but it also might be related to unfortunate side-effects of shortcomings within the test suite, as only three test cases are failing. Let me bring in this patch first, and then have another look after rebasing.
> 
>     * [Dialect: Fix `get_table_names()` reflection method sqlalchemy-cratedb#10](https://github.com/crate-workbench/sqlalchemy-cratedb/pull/10)

Yes I think it's unrelated. This PR should not have issues anymore now.

amotl (approved):
Thank you!

tests/util/database.py
```diff
@@ -0,0 +1,12 @@
...
+
+
+def test_quote_ident():
+    database = DatabaseAdapter(dburi="crate://localhost")
+    assert database.quote_ident("my_table") == "my_table"
+    assert database.quote_ident("my-table") == '"my-table"'
+    assert database.quote_ident("MyTable") == '"MyTable"'
+    assert database.quote_ident('"MyTable"') == '"MyTable"'
+    assert database.quote_ident("my_schema.my_table") == "my_schema.my_table"
+    assert database.quote_ident("my-schema.my_table") == '"my-schema".my_table'
```
amotl, line 11: What about this case and friends?
```python
assert database.quote_ident("my_schema"."my_table") == "my_schema.my_table"
```

seut: The current logic will not remove such valid quotes, why should it?

seut:
I think we should rename the `quote_ident` function to `quote_relation_name` to express it's difference from DB's usual `quote_ident` functions.
As written at https://github.com/crate/sqlalchemy-cratedb/issues/69#issuecomment-2167394339, a `quote` or `quote_identifier` function is generally not able and intended to work on full-qualified (relation/column) names but only on the unqualified identifier part.

seut:
@amotl 
I've added a sanity check at the quoting function and also renamed it to express better what it does, pls check.

amotl (approved):

Issue #148, opened 2024-05-20
[I/O] Add `--overwrite` option for `ctk load table` subsystem


amotl:
## About

@wierdvanderhaar reported that the InfluxDB I/O subsystem lacks an `--overwrite` option. Currently, the implementation probably always overwrites the target collection.

amotl:
Currently, the process always overwrites the table by dropping and recreating it. That's unfortunate.

amotl:
Optimally forward `--overwrite` to Dask's / pandas' `if_exists` parameter.
```python
if_exists : {'fail', 'replace', 'append'}
```

amotl:
The situation has been improved. Relevant updates will be part of the next release.

- https://github.com/daq-tools/influxio/pull/129
- https://github.com/crate-workbench/cratedb-toolkit/pull/176

PR #176, opened 2024-06-23
IO: Add the `if-exists` query parameter by updating to influxio 0.4.0

amotl:
## Problem
Because the InfluxDB I/O subsystem, implementing `ctk load table influxdb2://...`, always replaced the target table, it could have been considered as "just a toy". Thanks to that discovery by @wierdvanderhaar, we have been able to improve it.

## Solution
The improvement follows the semantics of the [`to_sql()`](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.to_sql.html) implementation of pandas/Dask, and now takes a new `if-exists` URL query parameter into consideration, forwarding the possible values `{fail,replace,append}` to the corresponding implementation.

## References
- GH-148
- https://github.com/daq-tools/influxio/pull/129

/cc @matkuliak, @simonprickett

### DIFF
--- a/doc/io/influxdb/loader.md
+++ b/doc/io/influxdb/loader.md
@@ -16,7 +16,7 @@ working with InfluxDB.
 pip install --upgrade 'cratedb-toolkit[influxdb]'
 ```
 
-## Examples
+## Usage
 
 ### Workstation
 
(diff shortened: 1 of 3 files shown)


doc/io/influxdb/loader.md
```diff
@@ -66,5 +66,26 @@ ctk load table \
   --cratedb-sqlalchemy-url="crate://admin:dZ...6LqB@green-shaak-ti.eks1.eu-west-1.aws.cratedb.net:4200/testdrive/demo?ssl=true"
 ```
 
+## Parameters
+
+### `if-exists`
+
+The target table will be created automatically, if it does not exist. The
+`if-exists` URL query parameter can be used to configure this behavior if it
+does. The default value is `fail`, the possible values are:
```
hammerhead, line 75:
```suggestion
The target table will be created automatically, if it does not exist. If it
does exist, the `if-exists` URL query parameter can be used to configure this
behavior. The default value is `fail`, the possible values are:
```

Wasn't clear to me with the "if it does" at the end of the sentence what it referred to, had to read it a few times until I got the reference.

hammerhead (approved):

simonprickett (approved):

PR #175, opened 2024-06-18
Add support to publish cfr as a binary using PyInstaller

amotl:
## About
Use [PyInstaller](https://pyinstaller.org/) to build self-contained application bundles of the [CrateDB Cluster Flight Recorder (CFR)](https://cratedb-toolkit.readthedocs.io/cfr/).

- GH-160

## Preview
See section "Artifacts » Produced during runtime" on [GHA run #9826830191](https://github.com/crate-workbench/cratedb-toolkit/actions/runs/9826830191).

PR #192, opened 2024-07-07
CFR: Upload application bundle artifacts to GitHub Pages

amotl:
## Problem
Uploading on PRs to GitHub Workflow Artifacts is not enough, because it does not provide good UX for downloading purposes.

## About
This patch extends the GHA workflow to additionally upload to GitHub Pages.

## References
- GH-175

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+cratedb-cfr filter=lfs diff=lfs merge=lfs -text
+cratedb-cfr.exe filter=lfs diff=lfs merge=lfs -text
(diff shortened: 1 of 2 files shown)


amotl:
GitHub Pages is not a very good upload destination, because artifacts/resources will be added to the repository itself by default, into a branch called `gh-pages`. Because the total payload size already is 4x ~100 MB, this would blow up the repository size too much, and already gets rejected by default:

> ```
> remote: error: File tmp/linux-x64/cratedb-cfr is 126.43 MB; this exceeds GitHub's file size limit of 100.00 MB        
> remote: error: GH001: Large files detected. You may want to try Git Large File Storage - https://git-lfs.github.com./
> ```

-- https://github.com/crate-workbench/cratedb-toolkit/actions/runs/9823563563/job/27121581872#step:10:77

Issue #215, opened 2024-08-06
[MongoDB I/O] Propagation of `_id` field is lost


amotl:
## Problem
CrateDB can't accept column names beginning with underscores.

## Solution
Within `OBJECT` columns, fields with names starting with underscores are ok.

### Introduction
@hlcianfagna asked:
> Do you recall if the import procedure as implemented by the [MongoDB Table Loader](https://cratedb-toolkit.readthedocs.io/io/mongodb/loader.html) can handle an `_id` column at the first level?

@surister responded:
> https://github.com/crate/cratedb-toolkit/blob/0c59124e6625c95cf2956cf02de5e0da8bd6961b/cratedb_toolkit/io/mongodb/export.py#L86-L91

### Evaluation
Within this routine, `del d["_id"]` might not be applicable, because it drops valuable information. It is sad, because `_id` can easily be used as a primary key in a generic way for all MongoDB collections, if the dataset doesn't define any other.

### Suggestions
> In some cases people may have used it as a foreign key to reference specific records in other tables so maybe I would be inclined to import it and declare it as the PK.

> I'd put the entire record from MongoDB into a single `OBJECT` in CrateDB: Within `OBJECT` columns, fields with names starting with underscores are ok.

Thanks, @hlcianfagna!

amotl:
Hi. I think this problem has been addressed with the recently landed bugfix and improvement patch, by including the `_id` top-level column again, and renaming it to `id`. So, I am closing this.

- GH-216

Thank you very much for your report, and please re-open if you think the problem still exists. Otherwise, please enjoy the fixed package. 🙏

Issue #210, opened 2024-08-02
MongoDB export doesn't write things to stdout.


surister:
In https://github.com/crate/cratedb-toolkit/blob/b7b80104aea30b7e47a6dd1298edf80fb42bfbdb/cratedb_toolkit/io/mongodb/export.py#L105 we have default `sys.stdout.buffer` but the only usage in https://github.com/crate/cratedb-toolkit/blob/b7b80104aea30b7e47a6dd1298edf80fb42bfbdb/cratedb_toolkit/io/mongodb/core.py#L118 we pass a `BytesIO` object, so in reality we never throw the documents to the stdout. 

Meaning that as per explained in https://cratedb-toolkit.readthedocs.io/io/mongodb/migr8.html, we cannot `migr8 export ... > data.json`

If we dummy fix this, we still get the header in the file, any tool taking advantage of this should take this into consideration.

![image](https://github.com/user-attachments/assets/e35eb7f8-7be9-416d-af01-b0cad2b5bb93)

surister:
In fact, the first info line breaks the integration with `cr8 insert-json`, if we eliminate it it works, for simple documents.
![image](https://github.com/user-attachments/assets/62aebb09-7c50-445d-8d11-a983dc44a709)

amotl:
Thank you. Maybe this bug has been introduced while wrapping migr8 into `ctk load table`. We planned a documentation update, see:

- https://github.com/crate/cratedb-guide/issues/76

But also, if there is a bug in migr8 about the issue you are observing, we should fix it.

amotl:
Hi. I think this problem has been addressed with the recently landed bugfix and improvement patch, so I am closing this.

- GH-216

Thank you very much for your report, and please re-open if you think the problem still exists. Otherwise, please enjoy the fixed package. 🙏

PR #216, opened 2024-08-08
MongoDB: Various fixes and improvements

amotl:
## About
The venerable migr8 program did not receive much love to get the details right. This patch intends to catch up on this, in order to make [`ctk load table mongodb://...`](https://cratedb-toolkit.readthedocs.io/io/mongodb/loader.html) hold up to its promise of better [DWIM](https://en.wikipedia.org/wiki/DWIM).

## Preview
```shell
pip install --upgrade 'cratedb-toolkit[mongodb] @ git+https://github.com/crate/cratedb-toolkit.git@mongodb-better-1'
```

### Bugfixes
- [x] GH-210
- [x] Fix timestamp parsing on `migr8 export`.
  `Exception: Can't parse datetime string 2022-06-11T04:37:53.200Z`
- [x] Converge `_id` input field to `id` column instead of dropping it.
  GH-215

### Details
- MongoDB: Fix missing output on STDOUT for `migr8 export`
- MongoDB: Improve timestamp parsing by using `python-dateutil`
- MongoDB: Converge `_id` input field to `id` column instead of dropping it
- MongoDB: Make user interface use stderr, so stdout is for data only
- MongoDB: Make `migr8 extract` write to stdout by default
- MongoDB: Make `migr8 translate` read from stdin by default
- MongoDB: Improve user interface messages
- MongoDB: Strip single leading underscore character from all top-level fields
- MongoDB: Map OID types to CrateDB TEXT columns
- MongoDB: Make `migr8 extract` and `migr8 export` accept the `--limit` option
- MongoDB: Fix indentation in prettified SQL output of `migr8 translate`
- MongoDB: Add capability to give type hints and add transformations using `--transformation`, see [Zyp Transformations](https://commons-codec.readthedocs.io/zyp/).

### DIFF
--- a/cratedb_toolkit/io/mongodb/core.py
+++ b/cratedb_toolkit/io/mongodb/core.py
@@ -38,7 +38,7 @@ def gather_collections(database) -> t.List[str]:
 
     rich.print(tbl)
 
-    rich.print("\nCollections to exclude: (eg: '0 1 2', '0, 1, 2', '0-2')")
+    rich.print("\nCollections to exclude: (eg: '0 1 2', '0, 1, 2', '0-2'). Leave empty for using all connections.")
 
     sys.stderr.write("> ")
     collections_to_ignore = parse_input_numbers(input())
--- a/cratedb_toolkit/io/mongodb/core.py
+++ b/cratedb_toolkit/io/mongodb/core.py
@@ -1,12 +1,14 @@
 import io
 import logging
+import sys
 import typing as t
 
 import pymongo
 import pymongo.database
-import rich
 from bson.raw_bson import RawBSONDocument
+from rich.console import Console
 from rich.syntax import Syntax
+from rich.table import Table
 
 from .export import collection_to_json
 from .extract import extract_schema_from_collection
@@ -14,6 +16,9 @@ from .translate import translate as translate_schema
 from .util import parse_input_numbers
 
 logger = logging.getLogger(__name__)
+
+console = Console(stderr=True)
+rich = console
 
 
 def gather_collections(database) -> t.List[str]:
@@ -23,7 +28,7 @@ def gather_collections(database) -> t.List[str]:
 
     collections = database.list_collection_names()
 
-    tbl = rich.table.Table(show_header=True, header_style="bold blue")
+    tbl = Table(show_header=True, header_style="bold blue")
     tbl.add_column("Id", width=3)
     tbl.add_column("Collection Name")
     tbl.add_column("Estimated Size")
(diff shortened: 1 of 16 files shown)


cratedb_toolkit/io/mongodb/core.py
```diff
@@ -33,9 +38,10 @@ def gather_collections(database) -> t.List[str]:
 
     rich.print(tbl)
 
-    rich.print("\nCollections to exclude: (eg: '0 1 2', '0, 1, 2', '0-2')")
+    rich.print("\nCollections to exclude: (eg: '0 1 2', '0, 1, 2', '0-2'). Leave empty for using all connections.")
```
amotl, line 41:
```suggestion
    rich.print("\nCollections to exclude: (eg: '0 1 2', '0, 1, 2', '0-2'). Leave empty to use all collections.")
```

cratedb_toolkit/io/mongodb/export.py
```diff
@@ -25,32 +25,20 @@
...
-        sign, hours, minutes = iso_match.groups()
-        tzoffset = int(hours) * 3600 + int(minutes) * 60
-        if sign == "-":
-            dt = dt + timedelta(seconds=tzoffset)
-        else:
-            dt = dt - timedelta(seconds=tzoffset)
-    else:
-        raise Exception("Can't parse datetime string {0}".format(value))
+    dt = dateparser.parse(value)
     return calendar.timegm(dt.utctimetuple()) * 1000
```
amotl, line 42: This has been trimmed to just rely on [dateutil.parser](https://dateutil.readthedocs.io/en/stable/parser.html) completely. Do you think it will be okay, because that code here just tried to achieve similar things manually?

hlcianfagna (commented):
Just a small thing that it may be worth mentioning in the docs, the schema translation part takes the name of the MongoDB collection for the `CREATE TABLE` statement but the data load part takes the table name from the `CRATEDB_SQLALCHEMY_URL` so these 2 have to match for `ctk load table` to work.

amotl:
Hi Hernan. Thanks for your suggestion. I was sure there are anomalies, but I can't spot them, probably because of operational blindness. Can I humbly ask you to submit a corresponding suggestion how and where to improve the documentation, either by commenting on this patch, or by submitting a separate one? Thanks a stack!

amotl:
The MongoDB Table Loader will now also accept Zyp transformations for re-shaping data.

- https://cratedb-toolkit--216.org.readthedocs.build/io/mongodb/loader.html#using-zyp-transformations

surister (approved):

PR #224, opened 2024-08-16
Dependencies: Update to lorrystream 0.0.4 and commons-codec 0.0.6. Adjust code. Add software tests.

amotl:
## Problem
On the last occasion, dependencies to lorrystream have not been verified correctly, so it was expectable that things go south at some point. Pinning to a branch, or pinning to a specific commit, are both not satisfying solutions to compensate properly for aligning things that are in flux by nature. C'est la vie.

## Solution
This patch fixes them, and adds rudimentary software tests to avoid tripping into the same situation as before.

## References
- https://github.com/crate/cratedb-toolkit/issues/222

/cc @wierdvanderhaar, @surister, @ckurze

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -155,3 +155,68 @@ jobs:
         env_vars: OS,PYTHON
         name: codecov-umbrella
         fail_ci_if_error: true
+
+
+  tests-kinesis:
+
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: ["ubuntu-latest"]
+        python-version: ["3.9", "3.12"]
+
+    env:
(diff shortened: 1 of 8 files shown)


hlcianfagna (commented):
This builds and now I can succesfully
```python
from cratedb_toolkit.iac.aws import DynamoDBKinesisPipe, LambdaFactory, LambdaPythonImage
```
but I cannot use the new `commons-codec` with it
```
lorrystream 0.0.3 requires commons-codec==0.0.3, but you have commons-codec 0.0.5 which is incompatible
```

amotl:
That other patch relaxes dependency constraints a bit. It will hopefully improve situations like that.

- https://github.com/daq-tools/lorrystream/pull/138

PR #226, opened 2024-08-19
DynamoDB: Add table loader for full-load operations

amotl:
## About
Bring DynamoDB full-load to Toolkit's `ctk load table` interface. 

## Documentation
https://cratedb-toolkit--226.org.readthedocs.build/io/dynamodb/loader.html

## Status
Alpha. For now, the implementation uses the same easy strategy to converge the source record into a single `data (OBJECT)` column in CrateDB. The 1:1 strategy may follow.

## Backlog
- GH-231

/cc @hammerhead, @zolbatar

### DIFF
--- a/cratedb_toolkit/io/dynamodb/backlog.md
+++ b/cratedb_toolkit/io/dynamodb/backlog.md
@@ -1,0 +1,35 @@
+# DynamoDB Backlog
+
+## Iteration +1
+- Pagination / Batch Getting.
+  https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/programming-with-python.html#programming-with-python-pagination
+
+- Use `batch_get_item`.
+  https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/dynamodb/client/batch_get_item.html
+
+- Scan by query instead of full.
+
+
+## Iteration +2
+
+### Resumption on errors?
+Another variant to scan the table, probably for resuming on errors?
+```python
+key = None
+while True:
+  if key is None:
+    response = table.scan()
+  else:
+    response = table.scan(ExclusiveStartKey=key)
+  key = response.get("LastEvaluatedKey", None)
+```
+
+### Item transformations?
+That's another item transformation idea picked up from an example program.
+Please advise if this is sensible in all situations, or if it's just a
+special case.
+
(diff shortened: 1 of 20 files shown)


cratedb_toolkit/io/dynamodb/copy.py
```diff
@@ -0,0 +1,97 @@
...
+            logger.info(f"Number of records written: {records_out}")
+            if records_out < records_in:
+                logger.warning("No data has been copied")
+
+    def items_to_sql(self, items):
+        """
+        Convert data for record items to INSERT statements.
+        """
+        for item in items:
+            yield self.translator.to_sql(item)
```
amotl, line 75: That's another item transformation idea picked up from an example program. Please advise if this is sensible in all situations, or if it's just a special case.

```python
if 'id' in item and not isinstance(item['id'], str):
    item['id'] = str(item['id'])
```

/cc @wierdvanderhaar

cratedb_toolkit/io/dynamodb/backlog.md
```diff
@@ -0,0 +1,9 @@
+# DynamoDB Backlog
+
+- Pagination / Batch Getting.
+  https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/programming-with-python.html#programming-with-python-pagination
+
+- Use `batch_get_item`.
+  https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/dynamodb/client/batch_get_item.html
+
+- Scan by query instead of full.
```
amotl, line 9: @wierdvanderhaar: With respect to alternative implementations, using batched reading from DynamoDB is probably way to go when processing large amounts of data?

wierdvanderhaar: Yes, we should use the batched method. Let's use a default batch size of 100 but give the option to use different batch sizes.

amotl: Hi. I added the remnant items to the backlog.

- GH-231

Do you agree to merge and release it first, following the "first make it work, then make it fast|beautiful|robust" paradigm, in order to get it out as quickly as possible?

wierdvanderhaar (approved):

Issue #219, opened 2024-08-14
MongoDB: Cannot cast value `1563051934000` to type `integer`


matkuliak:
Hi. I've been trying out the `ctk load table` with MongoDB. Local instances work well, haven't encountered any issues there.
But I'm having some trouble with cloud-to-cloud variant. _TLDR is that when I use the tool, it only creates the table that should be copied, without copying the contents._ 

It creates it with correct schema, but the actual data isn't copied. Details below. It might be the case that it's a syntax issue on my part, but as the table is created I think it should be correct.

## Sample data in Atlas (MongoDB Cloud)
```3:12:34 ~ $ mongosh
Current Mongosh Log ID: 66bc048475d29f3dc9838725
Connecting to:          mongodb://127.0.0.1:27017/?directConnection=true&serverSelectionTimeoutMS=2000&appName=mongosh+2.2.15
Using MongoDB:          7.0.12
Using Mongosh:          2.2.15

For mongosh info see: https://docs.mongodb.com/mongodb-shell/

------
   The server generated these startup warnings when booting
   2024-08-13T05:57:43.700+00:00: Using the XFS filesystem is strongly recommended with the WiredTiger storage engine. See http://dochub.mongodb.org/core/prodnotes-filesystem
   2024-08-13T05:57:43.873+00:00: Access control is not enabled for the database. Read and write access to data and configuration is unrestricted
   2024-08-13T05:57:43.873+00:00: /sys/kernel/mm/transparent_hugepage/enabled is 'always'. We suggest setting it to 'never' in this binary version
   2024-08-13T05:57:43.873+00:00: vm.max_map_count is too low
------

test> show dbs
admin      40.00 KiB
config     72.00 KiB
local      40.00 KiB
testdrive  40.00 KiB
test> use testdrive
switched to db testdrive
testdrive> show collections
demo
testdrive> db.demo.find()
[
  {
    _id: ObjectId('66baf67b55e4abaa15cc8cba'),
    VendorID: 2,
    tpep_pickup_datetime: Long('1563051934000'),
    tpep_dropoff_datetime: Long('1563053222000'),
    passenger_count: 2,
    trip_distance: 3.29,
    RatecodeID: 1,
    store_and_fwd_flag: 'N',
    PULocationID: 79,
    DOLocationID: 170,
    payment_type: 1,
    fare_amount: 15.5,
    extra: 0.5,
    mta_tax: 0.5,
    tip_amount: 3.86,
[...]

PR #235, opened 2024-08-21
MongoDB: Improve type mapper discriminating between INTEGER and BIGINT

amotl:
## About
@matkuliak reported that the MongoDB I/O subsystem wasn't capable to map values like `Long('1563051934000')` to CrateDB SQL's `BIGINT` data type.

- GH-219

This patch improves the situation.

### DIFF
--- a/cratedb_toolkit/io/mongodb/extract.py
+++ b/cratedb_toolkit/io/mongodb/extract.py
@@ -182,5 +182,17 @@ TYPES_MAP = {
 }
 
 
-def get_type(o):
-    return TYPES_MAP.get(type(o), "UNKNOWN")
+def get_type(value):
+    """
+    Resolve value type via type map, with special treatment for integer types.
+
+    INTEGER: -2^31 to 2^31-1
+    BIGINT: -2^63 to 2^63-1
+    """
+    type_ = type(value)
+    if type_ is int:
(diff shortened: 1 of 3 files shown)


cratedb_toolkit/io/mongodb/extract.py
```diff
@@ -182,5 +182,17 @@ def extract_schema_from_array(array: list, schema: dict):
...
-def get_type(o):
-    return TYPES_MAP.get(type(o), "UNKNOWN")
+def get_type(value):
+    """
+    Resolve value type via type map, with special treatment for integer types.
+
+    INTEGER: -2^31 to 2^31-1
+    BIGINT: -2^63 to 2^63-1
+    """
+    type_ = type(value)
```
surister, line 192: `isinstance(value, int)` isn't this better?

amotl: `type_` is used to look up the target type from the defined map in `TYPES_MAP`.

https://github.com/crate/cratedb-toolkit/blob/d1c0740a218bba94dfe70d0e8786b4fdc5d4f1f0/cratedb_toolkit/io/mongodb/extract.py#L168-L182

surister: right

PR #243, opened 2024-08-26
DMS/DynamoDB/MongoDB: Use SQL with parameters instead of inlining values

amotl:
## About
This patch adjusts for an improvement to commons-codec where the `SQLOperation` class has been introduced to bundle together an SQL statement and its parameters.

## References
- https://github.com/crate/commons-codec/pull/35


tests/io/test_processor.py
```diff
@@ -33,3 +109,66 @@ def test_processor_invoke_no_records(reset_handler, mocker, caplog):
...
+        "Records": [
+            wrap_kinesis(DYNAMODB_CDC_MODIFY_NESTED),
+        ]
+    }
+
+    # Run transfer command.
+    handler(event1, None)
+    cratedb.database.refresh_table(table_name)
+    handler(event2, None)
+    cratedb.database.refresh_table(table_name)
```
amotl, line 153: By adding and exercising this test case, I observed that an INSERT operation followed by an UPDATE operation on the same document/record indeed requires a REFRESH operation in between.

cratedb_toolkit/io/processor/kinesis_lambda.py
```diff
@@ -123,8 +123,13 @@ def handler(event, context):
...
             # Process record.
-            sql = cdc.to_sql(record_data)
-            connection.execute(sa.text(sql))
+            operation = cdc.to_sql(record_data)
+            connection.execute(sa.text(operation.statement), parameters=operation.parameters)
+
+            # Processing alternating CDC events requires write synchronization.
+            # FIXME: Needs proper table name quoting.
+            connection.execute(sa.text(f"REFRESH TABLE {CRATEDB_TABLE}"))
+
```
amotl, line 132: Hereby, I am adding a REFRESH TABLE operation again. Please share your objections if you have any.

/cc @wierdvanderhaar, @hlcianfagna, @hammerhead, @proddata, @widmogrod

tests/io/test_processor.py
amotl: This section has been amended in favor of adding the REFRESH TABLE operation to the handler.

pyproject.toml
```diff
@@ -139,7 +139,7 @@ docs = [
 ]
 dynamodb = [
   "boto3",
-  "commons-codec",
+  "commons-codec @ git+https://github.com/crate/commons-codec.git@sql-with-params",
```
amotl, line 142: Sure enough, this and other relevant spots need to be amended after running another release of commons-codec.

PR #245, opened 2024-08-27
DMS/DynamoDB: Fix table name quoting within CDC processor handler

amotl:
## About
Easily fix missing table quoting within CDC processor for DMS and DynamoDB events. It will need more improvements by adding the canonical implementation [quote_relation_name](https://github.com/crate/cratedb-toolkit/blob/v0.0.19/cratedb_toolkit/util/database.py#L44-L80) to sqlalchemy-cratedb, but may be good for now.

### DIFF
--- a/cratedb_toolkit/io/processor/kinesis_lambda.py
+++ b/cratedb_toolkit/io/processor/kinesis_lambda.py
@@ -127,8 +127,7 @@ def handler(event, context):
             connection.execute(sa.text(operation.statement), parameters=operation.parameters)
 
             # Processing alternating CDC events requires write synchronization.
-            # FIXME: Needs proper table name quoting.
-            connection.execute(sa.text(f"REFRESH TABLE {CRATEDB_TABLE}"))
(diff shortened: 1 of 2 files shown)


cratedb_toolkit/io/processor/kinesis_lambda.py
```diff
@@ -127,8 +127,7 @@ def handler(event, context):
             connection.execute(sa.text(operation.statement), parameters=operation.parameters)
 
             # Processing alternating CDC events requires write synchronization.
-            # FIXME: Needs proper table name quoting.
-            connection.execute(sa.text(f"REFRESH TABLE {CRATEDB_TABLE}"))
+            connection.execute(sa.text(f"REFRESH TABLE {cdc.quote_table_name(CRATEDB_TABLE)}"))
```
amotl, line 130: See [quote_table_name](https://github.com/crate/commons-codec/blob/v0.0.13/src/commons_codec/transform/dynamodb.py#L61-L70). It is currently just a poor man's version until we can shuffle code better and more sustainably.

wierdvanderhaar (approved):
Lets go!

PR #255, opened 2024-09-02
MongoDB: More ad hoc fixes for supporting real-world data

amotl:
## About
- Add support for UUID types.
- Improve reading timestamps in previous BSON formats like `end_date: Long('1455141600000')`.
- Make `ctk load table` use the `data OBJECT(DYNAMIC)` mapping strategy.
- Add relevant end-to-end integration test.
- Sanitize lists of varying objects.
- ~~Add `--treatment` option for applying special treatments to certain items on real-world data.~~
- Improve Zyp Transformations (`--transformation` option) for applying special treatments to certain items on real-world data.
- Use pagination on source collection, for creating batches towards CrateDB.
- Unlock importing MongoDB Extended JSON files using `file+bson://...`.

## Documentation
https://cratedb-toolkit--255.org.readthedocs.build/io/mongodb/loader.html

## Install
```
pip install --upgrade 'cratedb-toolkit[mongodb] @ git+https://github.com/crate/cratedb-toolkit.git@mongodb-more'
```

## Backlog
- [x] Add relevant software tests.
- [x] Documentation. For example, parameter `batch-size`.

/cc @hlcianfagna, @hammerhead, @lservini, @zolbatar, @juanpardo, @widmogrod

PR #261, opened 2024-09-11
MongoDB: Improve support for reading JSON/BSON files

amotl:
## About
A few features that have not been ready (lacked software tests) to be included into GH-255.

## Details
- Unlock processing multiple collections, either from server database, or from filesystem directory
- Unlock processing JSON files from HTTP resource, using `https+bson://`
- Optionally filter server collection using MongoDB query expression

## Documentation
https://cratedb-toolkit--261.org.readthedocs.build/io/mongodb/loader.html

## Install
```
pip install --upgrade 'cratedb-toolkit[mongodb] @ git+https://github.com/crate/cratedb-toolkit.git@amo/mongodb-more-2'
```

### DIFF
--- a/cratedb_toolkit/api/main.py
+++ b/cratedb_toolkit/api/main.py
@@ -119,14 +119,15 @@ class StandaloneCluster(ClusterBase):
         if source_url_obj.scheme.startswith("dynamodb"):
             from cratedb_toolkit.io.dynamodb.api import dynamodb_copy
 
-            if not dynamodb_copy(str(source_url_obj), target_url, progress=True):
-                msg = "Data loading failed"
-                logger.error(msg)
-                raise OperationFailed(msg)
+            if dynamodb_copy(str(source_url_obj), target_url, progress=True):
+                return True
+            else:
+                logger.error("Data loading failed or incomplete")
+                return False
 
         elif source_url_obj.scheme.startswith("file"):
             if "+bson" in source_url_obj.scheme or "+mongodb" in source_url_obj.scheme:
-                mongodb_copy_generic(
+                return mongodb_copy_generic(
                     str(source_url_obj),
                     target_url,
                     transformation=transformation,
@@ -135,7 +136,7 @@ class StandaloneCluster(ClusterBase):
 
         elif source_url_obj.scheme.startswith("http"):
             if "+bson" in source_url_obj.scheme or "+mongodb" in source_url_obj.scheme:
-                mongodb_copy_generic(
+                return mongodb_copy_generic(
                     str(source_url_obj),
                     target_url,
                     transformation=transformation,
@@ -149,19 +150,20 @@ class StandaloneCluster(ClusterBase):
             if asbool(source_url_obj.query_params.get("ssl")):
                 http_scheme = "https"
             source_url_obj.scheme = source_url_obj.scheme.replace("influxdb2", http_scheme)
-            if not influxdb_copy(str(source_url_obj), target_url, progress=True):
-                msg = "Data loading failed"
-                logger.error(msg)
-                raise OperationFailed(msg)
+            if influxdb_copy(str(source_url_obj), target_url, progress=True):
+                return True
+            else:
+                logger.error("Data loading failed or incomplete")
+                return False
 
         elif source_url_obj.scheme.startswith("mongodb"):
             if "+cdc" in source_url_obj.scheme:
                 source_url_obj.scheme = source_url_obj.scheme.replace("+cdc", "")
                 from cratedb_toolkit.io.mongodb.api import mongodb_relay_cdc
 
-                mongodb_relay_cdc(str(source_url_obj), target_url, progress=True)
+                return mongodb_relay_cdc(str(source_url_obj), target_url, progress=True)
             else:
-                mongodb_copy_generic(
+                return mongodb_copy_generic(
                     str(source_url_obj),
                     target_url,
                     transformation=transformation,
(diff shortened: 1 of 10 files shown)


bmunkholm:
@cla-bot check

cratedb_toolkit/api/main.py
```diff
@@ -119,14 +119,24 @@ def load_table(self, resource: InputOutputResource, target: TableAddress, transf
...
             if "+bson" in source_url_obj.scheme or "+mongodb" in source_url_obj.scheme:
-                mongodb_copy_generic(
+                return mongodb_copy_generic(
+                    str(source_url_obj),
+                    target_url,
+                    transformation=transformation,
+                    progress=True,
+                )
+
+        elif source_url_obj.scheme.startswith("http"):
```
seut, line 137: Maybe move this up to `if source_url_obj.scheme.startswith("file") or source_url_obj.scheme.startswith("http")` as the body of both branches looks the same?

cratedb_toolkit/io/mongodb/api.py
```diff
@@ -96,22 +100,74 @@ def mongodb_copy(source_url, target_url, transformation: t.Union[Path, None] = N
...
+        logger.info(f"Discovered collections: {len(collections)}")
+        logger.debug(f"Processing collections: {collections}")
+        for collection_path in collections:
+            mongodb_uri_effective = mongodb_uri.navigate(Path(collection_path).name)
+            mongodb_uri_effective.query_params = mongodb_query_parameters
+            cratedb_uri_effective = cratedb_uri.navigate(Path(collection_path).stem)
+            tasks.append(
+                MongoDBFullLoad(
+                    mongodb_url=str(mongodb_uri_effective),
+                    cratedb_url=str(cratedb_uri_effective),
```
seut, line 155: Is the URL->String conversion here needed as the MongoDBFullLoad will convert the String back to an URL anyhow?

cratedb_toolkit/api/main.py
amotl: Significantly improved with fb70a36953. Thanks!

cratedb_toolkit/io/mongodb/api.py
amotl: Improved with 6bbda4da44. Thanks again!

cratedb_toolkit/io/mongodb/api.py
```diff
@@ -135,32 +139,26 @@ def mongodb_copy(
...
-            mongodb_uri.scheme.startswith("mongodb")
-            and Path(mongodb_uri.path).is_absolute()
-            and mongodb_uri.path[-1] != "/"
-        ):
-            mongodb_uri.path += "/"
-        if cratedb_uri.path[-1] != "/":
-            cratedb_uri.path += "/"
-        mongodb_query_parameters = mongodb_uri.query_params
-        mongodb_adapter = mongodb_adapter_factory(mongodb_uri)
+        address_pair_root = AddressPair(source_url=source_url, target_url=target_url)
```
amotl, line 146: This hackery has now also been refactored and wrapped away into an `AddressPair` object, in order to separate concerns...

cratedb_toolkit/io/mongodb/api.py
```diff
@@ -135,32 +139,26 @@ def mongodb_copy(
...
+            address_pair = address_pair_root.navigate(
+                source_path=Path(collection_path).name,
+                target_path=Path(collection_path).stem,
+            )
             tasks.append(
                 MongoDBFullLoad(
-                    mongodb_url=mongodb_uri_effective,
-                    cratedb_url=cratedb_uri_effective,
+                    mongodb_url=address_pair.source_url,
+                    cratedb_url=address_pair.target_url,
```
amotl, line 161: ... and to minimize its API surface. Now, you are just invoking `.navigate()` on that composite object instance and it will adjust its managed URL instances correspondingly.

cratedb_toolkit/model.py
```diff
@@ -120,3 +123,44 @@ class InputOutputResource:
...
+    target_url: URL
+
+    _source_url_query_parameters: t.Dict[str, t.Any] = Factory(dict)
+    _target_url_query_parameters: t.Dict[str, t.Any] = Factory(dict)
+
+    __SERVER_SCHEMES__ = ["http", "https", "mongodb", "mongodb+srv"]
+
+    def navigate(self, source_path: str, target_path: str) -> "AddressPair":
+        source_url_query_parameters = self.source_url.query_params
+        target_url_query_parameters = self.target_url.query_params
```
seut, line 145: As the query params are not changed in this method, why do we need to store (and copy) them separately?

cratedb_toolkit/model.py
```diff
@@ -120,3 +123,44 @@ class InputOutputResource:
...
+
+        source_url = URL(str(self.source_url))
```
seut, line 147: I guess there is no better API at the URL class to avoid string generation + parsing only to copy the instance, right?

cratedb_toolkit/model.py
amotl: The fundamental `.navigate()` method on the URL object implicitly gets rid of them, but we want to propagate them, so we forward them explicitly.

cratedb_toolkit/model.py
amotl: It looks like standard `deepcopy` is just fine, as already applied on other spots in this file. Improved with 2a5df04467b.

cratedb_toolkit/model.py
amotl: 94b52a8c143 stops using the `.navigate()` method of the designated URL library, and uses `urljoin` instead to compute and directly set the `.path` attribute. The result looks more streamlined and compact than before. Thanks!

seut (approved):
👍 thx

PR #262, opened 2024-09-13
MongoDB: Improve error handling wrt. bulk operations vs. usability

amotl:
## About
In order to have both, efficient bulk insert operations, and on-the-spot error messages on records that fail to insert, let's introduce a two-stage approach:

- First, try to insert a batch.
- When it fails, determine invalid records and insert them one-by-one, in order to relay corresponding error messages to the user.

## Details
`BulkProcessor` rips the canonical implementation off the specific handler about MongoDB I/O, setting the stage to reuse it on other I/O subsystems as well. Improving error handling has already been requested for the DynamoDB table loader, so applying the outcome there will certainly be next.

## References
- Related: https://github.com/crate/crate/issues/16617
- Predecessor: GH-261

/cc @juanpardo

### DIFF
--- a/cratedb_toolkit/io/core.py
+++ b/cratedb_toolkit/io/core.py
@@ -1,0 +1,180 @@
+# TODO: Maybe refactor to `sqlalchemy-cratedb` or `commons-codec` on another iteration?
+import typing as t
+from functools import cached_property
+
+import sqlalchemy as sa
+from attr import Factory
+from attrs import define
+from commons_codec.model import SQLOperation
+from pympler.asizeof import asizeof
+from sqlalchemy.exc import ProgrammingError
+from tqdm import tqdm
+
+from cratedb_toolkit.util.database import logger
+
+
+class BulkResultItem(t.TypedDict):
+    """
+    Define the shape of a CrateDB bulk request response item.
+    """
+
+    rowcount: int
+
+
+@define
+class BulkResponse:
+    """
+    Manage CrateDB bulk request responses.
+    Accepts a list of bulk arguments (parameter list) and a list of bulk response items.
+
+    https://cratedb.com/docs/crate/reference/en/latest/interfaces/http.html#bulk-operations
+
+    TODO: Think about refactoring this to `sqlalchemy_cratedb.support`.
+    """
+
+    parameters: t.Union[t.List[t.Dict[str, t.Any]], None]
+    cratedb_bulk_result: t.Union[t.List[BulkResultItem], None]
+
+    @cached_property
+    def failed_records(self) -> t.List[t.Dict[str, t.Any]]:
+        """
+        Compute list of failed records.
+
+        CrateDB signals failed insert using `rowcount=-2`.
+
+        https://cratedb.com/docs/crate/reference/en/latest/interfaces/http.html#error-handling
+        """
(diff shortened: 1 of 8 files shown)


cratedb_toolkit/sqlalchemy/patch.py
```diff
@@ -0,0 +1,19 @@
...
+    result = cursor.executemany(statement, parameters)
+    if context is not None:
+        context.last_executemany_result = result
+
+
+def monkeypatch_executemany():
+    """
+    Enable improved version of `do_executemany`.
+    """
+    dialect.do_executemany = do_executemany
```
amotl, line 19: That code needs to go into the SQLAlchemy dialect package on a later iteration. It sets the stage to relay [CrateDB's bulk operations response](https://cratedb.com/docs/crate/reference/en/latest/interfaces/http.html#bulk-operations) through SQLAlchemy's `ExecutionContext` to downstream consumers.

cratedb_toolkit/io/core.py
```diff
@@ -0,0 +1,180 @@
...
+            self.progress_bar and self.progress_bar.set_description("SUBMIT ")
+            try:
+                cursor = self.connection.execute(statement=statement, parameters=operation.parameters)
+                self.connection.commit()
+                cratedb_bulk_result = getattr(cursor.context, "last_executemany_result", None)
+                bulk_response = BulkResponse(operation.parameters, cratedb_bulk_result)
+                failed_records = bulk_response.failed_records
+                count_success_local = bulk_response.success_count
+                self._metrics.count_success_total += bulk_response.success_count
+                self.progress_bar and self.progress_bar.update(n=bulk_response.success_count)
```
amotl, line 149: This is the spot where the bulk operations response is being consumed and evaluated, by correlating the ingress parameters (batch of records) against the bulk ops response's status items `{"rowcount": 1}` vs. `{"rowcount": -2}`, in order to discover the effective records that failed.

seut (commented):
Nice idea, looks goot but isn't this missing a test case? Or how do you ensure that it's working like expected?

tests/io/mongodb/test_copy.py
```diff
@@ -187,6 +187,26 @@ def test_mongodb_copy_filesystem_json_relaxed(caplog, cratedb):
...
+
+    # Run transfer command.
+    mongodb_copy(json_resource, cratedb_url)
+
+    # Verify metadata in target database.
+    assert cratedb.database.table_exists("testdrive.demo") is True
+    assert cratedb.database.refresh_table("testdrive.demo") is True
+    assert cratedb.database.count_records("testdrive.demo") == 1
+
+    assert "Dynamic nested arrays are not supported" in caplog.text
```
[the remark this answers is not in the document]
amotl: I've just amended the patch so the `mixed.ndjson` file include three records, where a single one is invalid, like this:
```json
{"_id":1,"name":"Foo","date":{"$date":"2011-01-14T08:00:00Z"}}
{"_id":2,"name":"Bar","date":{"$date":"2011-01-15T08:00:00Z"},"nested_array":[[1,2]]}
{"_id":3,"name":"Baz","date":{"$date":"2011-01-16T08:00:00Z"}}
```

That aims to validate that processing does not stop on invalid records.

seut (approved):

PR #278, opened 2024-09-26
MongoDB: Update adapter to pymongo 4.9

amotl:
pymongo 4.9 has been released, and includes a few breaking API changes after adding an asynchronous variant to the driver.


amotl:
The [PyMongo Adapter for CrateDB](https://cratedb-toolkit.readthedocs.io/adapter/pymongo.html) was merely an exercise, but not a real product, so it will not receive significant maintenance any longer. However, it works well in the current status quo with the [dependencies as defined in the corresponding `pymongo` extra requirements section](https://cratedb-toolkit.readthedocs.io/adapter/pymongo.html#usage), so the code doesn't necessarily need to be removed from the code base unless there are other objections.

@florinutz or @kneth: I am closing this. Please delete the branch associated with this PR. Thank you!

PR #281, opened 2024-09-30
DynamoDB: Change CrateDB data model to use (`pk`, `data`, `aux`) columns

amotl:
## About
By breaking the primary key information out of the main record's data bucket, the main record can be updated as-is on CDC MODIFY operations.

## References
- https://github.com/crate/commons-codec/pull/64


cratedb_toolkit/io/dynamodb/copy.py
```diff
@@ -54,12 +53,16 @@ def start(self):
         """
         Read items from DynamoDB table, convert to SQL INSERT statements, and submit to CrateDB.
         """
-        records_in = self.dynamodb_adapter.count_records(self.dynamodb_table)
+        records_in = self.dynamodb_adapter.count_records(table_name=self.dynamodb_table)
         logger.info(f"Source: DynamoDB table={self.dynamodb_table} count={records_in}")
 
+        # FIXME!
```
hammerhead, line 59: Do we need a comment here on what is open to be fixed? At least I wouldn't know.

hammerhead (approved):

Issue #171, opened 2024-06-13
[wtf] Provide `tail -f` on top of `sys.jobs_log` table


amotl:
## Feature Pitch
> I would like to implement tailing the `sys.jobs_log` table in one way or another. Actually, `tail --follow`.

## Status
- Deferred from https://github.com/crate-workbench/cratedb-toolkit/pull/88#discussion_r1412119817.

amotl:
A first variant has been implemented.

- GH-330

amotl:
### Conclusion

Implementation and [documentation](https://cratedb-toolkit.readthedocs.io/util/tail.html) exist today, and I am avidly using the tool to investigate database conversations, enjoying nicely pretty-formatted SQL statements at your fingertips, ready for copy/paste into other documents.

/cc @karynzv, @hammerhead, @zolbatar, @grbade, @matriv, @WalBeh, @seut 

### Synopsis
```shell
docker run --rm --publish=4200:4200 crate '-Cdiscovery.type=single-node'
export CRATEDB_CLUSTER_URL="crate://"
uvx --from=cratedb-toolkit ctk tail -n 3 --follow --format=log-pretty sys.jobs_log
```

<img width="403" height="277" alt="Image" src="https://github.com/user-attachments/assets/89b3edc8-0ffa-4f41-8ecc-39bbc14a1344" />

amotl:
- Converged into documentation improvements with bbe1b831d6 in GH-779.

PR #330, opened 2024-12-16
Add basic utility command `ctk tail`, for tailing a database table and optionally following the tail

amotl:
## About
Implement `tail -f` for `sys.jobs_log`, to support ad hoc tracing of SQL statements processed by CrateDB.

## Synopsis
```shell
export CRATEDB_SQLALCHEMY_URL=crate://crate@localhost:4200/
```
```shell
ctk tail -n 3 --follow --format=log-pretty sys.jobs_log
crash -c "SELECT * FROM sys.summits LIMIT 3;"
```
The implementation may be versatile enough to also work well on arbitrary tables.
```shell
ctk tail -n 3 sys.summits
```

## Documentation
- https://cratedb-toolkit--330.org.readthedocs.build/cmd/
- https://cratedb-toolkit--330.org.readthedocs.build/wtf/#one-shot-commands

## References
- GH-171

### DIFF
--- a/cratedb_toolkit/cmd/tail/cli.py
+++ b/cratedb_toolkit/cmd/tail/cli.py
@@ -22,6 +22,7 @@ cratedb_sqlalchemy_option = click.option(
 )
 @click.option("--format", "format_", type=str, required=False, help="Select output format. Default: log / jsonl")
 @click.option("--follow", "-f", is_flag=True, required=False, help="Follow new records added, by polling the table")
+@click.option("--interval", "-i", type=float, required=False, help="When following the tail, poll each N seconds. Default: 0.1")
 @click.option("--verbose", is_flag=True, required=False, help="Turn on logging")
 @click.option("--debug", is_flag=True, required=False, help="Turn on logging with debug level")
 @click.argument("resource", nargs=-1, type=click.UNPROCESSED)
@@ -34,6 +35,7 @@ def cli(
     lines: int,
     format_: str,
     follow: bool,
+    interval: float,
     verbose: bool,
     debug: bool,
 ):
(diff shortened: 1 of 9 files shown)


cratedb_toolkit/cmd/tail/cli.py
```diff
@@ -0,0 +1,52 @@
...
+
+@click.command()
+@cratedb_sqlalchemy_option
+@click.option(
+    "--lines", "-n", type=int, required=False, default=10, help="Displays n last lines of the input. Default: 10"
+)
+@click.option("--format", "format_", type=str, required=False, help="Select output format. Default: log / jsonl")
+@click.option("--follow", "-f", is_flag=True, required=False, help="Follow new records added, by polling the table")
+@click.option("--verbose", is_flag=True, required=False, help="Turn on logging")
+@click.option("--debug", is_flag=True, required=False, help="Turn on logging with debug level")
```
seut, line 26: Looks like the `--interval` option is not defined and passed through?

seut (commented):
Great stuff! Spotted a cli arg issue..

cratedb_toolkit/cmd/tail/cli.py
amotl: Nice spot. Thanks!

cratedb_toolkit/cmd/tail/cli.py
```diff
@@ -0,0 +1,56 @@
...
+@click.option(
+    "--lines", "-n", type=int, required=False, default=10, help="Displays n last lines of the input. Default: 10"
+)
+@click.option("--format", "format_", type=str, required=False, help="Select output format. Default: log / jsonl")
+@click.option("--follow", "-f", is_flag=True, required=False, help="Follow new records added, by polling the table")
+@click.option(
+    "--interval", "-i", type=float, required=False, help="When following the tail, poll each N seconds. Default: 0.5"
+)
+@click.option("--verbose", is_flag=True, required=False, help="Turn on logging")
+@click.option("--debug", is_flag=True, required=False, help="Turn on logging with debug level")
```
seut, line 29: Any reason not to use `click` default option settings for all default values instead of later on inside the code?

seut (approved):
👍

cratedb_toolkit/cmd/tail/cli.py
amotl: If the defaults would be applied to click, they wouldn't apply when using ctk as a Python API in your own programs.

PR #339, opened 2025-01-09
Replaced old references to cratedb-datasets repo with new CrateDB URLs.

simonprickett:
We can't trust GitHub for access to LFS files directly when running `COPY FROM` statements, so this changes the URLs to point to their replacements.


amotl (approved):
Thank you!

PR #360, opened 2025-01-30
Datasets: Register UK wind farm dataset

amotl:
## About
By providing a little API that wraps SQL loader snippets of [cratedb-datasets](https://github.com/crate/cratedb-datasets), it becomes easier to load the wind farms dataset programmatically.

## References
- https://github.com/crate/cratedb-datasets/pull/26

/cc @simonprickett

### DIFF
--- a/cratedb_toolkit/datasets/tutorial.py
+++ b/cratedb_toolkit/datasets/tutorial.py
@@ -36,8 +36,24 @@ weather = Dataset(
     init_includes_loading=True,
 )
 
+windfarm_uk_info = Dataset(
+    reference="tutorial/windfarm-uk-info",
+    data_url="https://cdn.crate.io/downloads/datasets/cratedb-datasets/devrel/uk-offshore-wind-farm-data/wind_farms.json",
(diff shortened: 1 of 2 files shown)


cratedb_toolkit/datasets/tutorial.py
```diff
@@ -36,8 +36,24 @@
     init_includes_loading=True,
 )
 
+windfarm_uk_info = Dataset(
+    reference="tutorial/windfarm-uk-info",
+    data_url="https://cdn.crate.io/downloads/datasets/cratedb-datasets/devrel/uk-offshore-wind-farm-data/wind_farms.json",
```
kneth, line 41: Should we declare `https://cdn.crate.io/downloads/datasets/cratedb-datasets` as a constant and use it all over?

kneth (approved):
My comment is meant as a suggestion for the future

cratedb_toolkit/datasets/tutorial.py
amotl: Good idea. I will change it on one of the future iteration. Thanks.

amotl: I've added e8b440f8 to GH-361.

PR #428, opened 2025-04-29
Anonymize sql statements

WalBeh:
## Summary of the changes
This allows to _anonymize_ the queries aggregated by `ctk cfr jobstats` by adding a `--anonymize` flag. Anonymization is built on the `https://queryanonymizer.com/` library. The strings replaced are saved to a json file (defaults to `decoder_dictionary.json`. This allows to reverse back to clear text, which might be useful for troubleshooting queries. 

```shell
ctk cfr jobstats collect --anonymize clusterABC.json
ctk cfr jobstats view --deanonymize clusterABC.json
```

### DIFF
--- a/cratedb_toolkit/cfr/jobstats.py
+++ b/cratedb_toolkit/cfr/jobstats.py
@@ -11,6 +11,7 @@ from uuid import uuid4
 
 import urllib3
 from crate import client
+from queryanonymizer import anonymize, deanonymize
 
 from cratedb_toolkit.model import DatabaseAddress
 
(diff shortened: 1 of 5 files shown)


amotl (approved):
Thanks!

cratedb_toolkit/cfr/jobstats.py
```diff
@@ -42,13 +43,40 @@
...
+        anonymize_sql, \
+        deanonymize_sql, \
+        decoder_dict_path
+    anonymize_sql = anonymize_statements
+    deanonymize_sql = deanonymize_statements
+    decoder_dict_path = decoder_dict_file
     schema = address.schema or "stats"
 
+    if anonymize_sql and decoder_dict_path:
+        logger.info(f"SQL anonymization is enabled, using dictionary: {decoder_dict_path}")
```
amotl, line 75: ## Problem
cratedb_toolkit/cfr/jobstats.py:71: error: Incompatible types in assignment (expression has type "str | None", variable has type "str")  [assignment]
-- https://github.com/crate/cratedb-toolkit/actions/runs/14751841553/job/41410919129?pr=428#step:6:53

## Evaluation
Because `decoder_dict_file` is declared optional, but `decoder_dict_path` possibly not, the `mypy` type checker raises an error here.

## Solution
You will need to guard spots like this like:
```python
if decoder_dict_file is None:
    raise ValueError("Could not pick up decoder file")
```

cratedb_toolkit/cfr/jobstats.py
```diff
@@ -42,13 +43,43 @@
...
+
+    if deanonymize_sql and decoder_dict_path:
+        logger.info(f"SQL deanonymization is enabled, using dictionary: {decoder_dict_path}")
+
+    if decoder_dict_file is None:
+        raise ValueError("Could not pick up decoder file")
```
amotl, line 81: Thanks for adding those guards in order to satisfy the mypy type checker. If you will move lines 80-81 to before line 71, it will signal success.

PR #458, opened 2025-06-20
Fix comparison of 0s vs 0ms

hlcianfagna:

Preview: `pip install --upgrade 'cratedb-toolkit[settings] @ git+https://github.com/crate/cratedb-toolkit.git@hlcianfagna/ctksettingscomparedefault0sdisabled'`

This addresses the issue of `stats.jobs_log_expiration` and `stats.operations_log_expiration` being flagged as not being default because the settings reads `0s` but the documentation says `0s (disabled)`

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## Unreleased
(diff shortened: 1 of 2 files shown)


amotl (approved):
Excellent, thanks!

Issue #443, opened 2025-06-04
`ctk load table` from `kinesis+dms` fails to parse ARN


hammerhead:
Straight from the [documentation](https://cratedb-toolkit.readthedocs.io/io/dms/standalone.html):

```bash
export CRATEDB_CLUSTER_URL=crate://crate@localhost:4200/testdrive/demo
ctk load table kinesis+dms://arn:aws:kinesis:eu-central-1:831394476016:stream/testdrive
```

Fails with:
```
Traceback (most recent call last):
  File "/Users/niklasschmidtmer/Documents/GitHub/.venv/lib/python3.13/site-packages/boltons/urlutils.py", line 932, in parse_url
    port = int(port_str)
ValueError: invalid literal for int() with base 10: 'aws:kinesis:eu-central-1:831394476016:stream'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/Users/niklasschmidtmer/Documents/GitHub/.venv/bin/ctk", line 8, in <module>
    sys.exit(cli())
             ~~~^^
  File "/Users/niklasschmidtmer/Documents/GitHub/.venv/lib/python3.13/site-packages/click/core.py", line 1161, in __call__
    return self.main(*args, **kwargs)
           ~~~~~~~~~^^^^^^^^^^^^^^^^^
  File "/Users/niklasschmidtmer/Documents/GitHub/.venv/lib/python3.13/site-packages/click/core.py", line 1082, in main
    rv = self.invoke(ctx)
  File "/Users/niklasschmidtmer/Documents/GitHub/.venv/lib/python3.13/site-packages/click/core.py", line 1697, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
                           ~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^
  File "/Users/niklasschmidtmer/Documents/GitHub/.venv/lib/python3.13/site-packages/click/core.py", line 1697, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
                           ~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^
  File "/Users/niklasschmidtmer/Documents/GitHub/.venv/lib/python3.13/site-packages/click/core.py", line 1443, in invoke
    return ctx.invoke(self.callback, **ctx.params)
           ~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #468, opened 2025-06-24
Kinesis: Improve argument parsing for stream ARNs

amotl:
## About
Support Kinesis stream ARNs with async-kinesis 2.x only.

## References
- GH-443

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 - Settings: Fixed comparison of `0s` vs `0ms`. Thanks, @hlcianfagna.
+- Kinesis: Improved argument parsing for stream ARNs. Thanks, @hammerhead.
 
 ## 2025/06/23 v0.0.36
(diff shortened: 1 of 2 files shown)


hammerhead (commented):
Could we consider bumping the dependency to `async-kinesis>2,<3 `? It would have the advantage of the documented example being guaranteed to work.

It is interesting that `pip3 install -U "cratedb-toolkit[kinesis]"` didn't update `async-kinesis` in my case. A manual `pip3 install -U async-kinesis` updated it to 2.2.0 and I can confirm `ctk load table` then initializes successfully with the original command.

amotl:
Hi Niklas.

> Could we consider bumping the dependency to `async-kinesis>2,<3`?

Sure we can also do that right away. Thanks.

> It did not work well for me when using `pip3 install -U <main package>`.

Yeah, that's sad. I also don't know exactly what the semantics of "--upgrade" are, in this case specifically related to the policy on transitive dependencies.

amotl:
Closing this in favor of GH-473. Thanks!

Issue #526, opened 2025-08-21
Next steps after XMover


amotl:
## Thoughts

After excellent tooling enhancements by @WalBeh per GH-523, we are thinking about the next steps how to improve operational tooling, in the spirit of an operators journey when fire-fighting a busy cluster.

## Topics

How to re-use existing components in ctk, mostly the cfr/info/jobstats subsystem, and how to glue them together.

- Report about hot shards @matriv will send SQL statements.
- Q: We want to know which tables are being queried with which queries. We do NOT want to know how long those statements take, the focus should be on the frequency of table use. Expected output:
  ```
  15:32:00 doc.Kochtopf (insert: 400, update: 20, select: 342)
  15:32:03 doc.Deckel (insert: 200, update: 10, select: 142)
  ```
    A: Report about hot tables (in terms of traffic/queries). Inquire from `sys.jobs_log` and do some stats on it. Caveat: When just looking at `sys.jobs_log`, you cannot directly identify tables when a query hits a view. Let's make the tool even smarter by considering the SQL VIEW DDL for backtracking.
- What did we do with cfr+marimo already?

matriv:
> @matriv will send SQL statements.
After looking carefullly into option what Juan has started with (seq numbers) is the way to go imho, ideally in combination with `sys.jobs_log` where we can filter out the write stmts like:
```
SELECT * FROM sys.jobs_log WHERE classification['type'] IN ('INSERT', 'UPDATE', 'DELETE') AND started BETWEEN ? AND ?;
```
and then (thx @surister) use something like:
```
from cratedb_sqlparse import sqlparse

t = sqlparse("""delete from table doc.test where v = 1""")

print(t[0].metadata.tables)
# [Table(name='test', schema='doc')]
```
to parse the stmt from the `sys.jobs_log` and get our hands on the table name.

amotl:
I did something the other day with this patch. It will need to be revisited and wrapped up, as it lost a bit of traction while the focus was on producing XMover.

- GH-528

PR #527, opened 2025-08-21
Admin/XMover: Shard heat monitoring, optionally watching

amotl:
## About

At GH-526, we are thinking about next steps in user ergonomics when debugging a busy cluster. Based on the idea of computing deltas of certain metrics available through CrateDB's system tables about shards, @juanpardo starts exploring a few ideas here. Thanks!

## Install
```shell
uv pip install --upgrade 'cratedb-toolkit @ git+https://github.com/crate/cratedb-toolkit.git@jp/xmover-shard-hot-monitor'
```

## Details

Static reports are always topmost priority. In an optional variant, a `--watch`-like interval-updating view could also make sense. However, it is difficult to learn about historic progress/changes in this case, i.e. when leaving the screen for a while, so discrete output is always preferred as the primary operation mode.

### DIFF
--- a/cratedb_toolkit/admin/xmover/cli.py
+++ b/cratedb_toolkit/admin/xmover/cli.py
@@ -13,11 +13,11 @@ from rich.console import Console
 from rich.panel import Panel
 
 from cratedb_toolkit.admin.xmover.analysis.shard import (
+    ActiveShardMonitor,
     ShardAnalyzer,
     ShardHeatReporter,
     ShardHeatSortByChoice,
     ShardReporter,
-    ActiveShardMonitor,
 )
 from cratedb_toolkit.admin.xmover.analysis.table import DistributionAnalyzer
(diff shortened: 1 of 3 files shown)


cratedb_toolkit/admin/xmover/cli.py
```diff
@@ -62,6 +62,16 @@ def analyze(ctx, table: Optional[str]):
...
 
 
+@main.command()
+@click.pass_context
+def monitor_shards(ctx):
+    """Monitor shards, pointing out hot ones"""
+    client = ctx.obj["client"]
+    analyzer = ShardAnalyzer(client)
+    monitor = ShardMonitor(analyzer)
+    monitor.monitor_shards()
```
amotl, line 72: Just an idea about »naming things«: `xmover monitor-shards` certainly fits the bill, but what do you think about `xmover shard-heat` [^1] to better reflect its intentions?

[^1]: ... with an optional `--watch` option in the long run?

PR #528, opened 2025-08-22
`ctk info table-traffic`

amotl:
## About

Find out about about table usage frequency.

> We want to know which tables are being queried with which queries.

## References

- GH-526


bgunebakan:
Hello, 
Thanks for taking the time to open this, and sorry it's been sitting for so long. We're doing a pass over older PRs, and this one has been inactive for a long time. if it's still relevant, please reopen and we'll take a look.

Best,
Bilal

PR #531, opened 2025-09-05
`ctk settings`: Workaround for SQLAlchemy sslmode error

amotl:
## About

A workaround for a yet unknown bug submitted by @tomach last week. Is it applicable or can we find a different one?

### DIFF
--- a/cratedb_toolkit/util/database.py
+++ b/cratedb_toolkit/util/database.py
@@ -8,9 +8,11 @@ import os
 import typing as t
 from pathlib import Path
 
+import requests
 import sqlalchemy as sa
 import sqlparse
 from boltons.urlutils import URL
+from crate import client as crate_client
 from cratedb_sqlparse import sqlparse as sqlparse_cratedb
 from sqlalchemy.exc import ProgrammingError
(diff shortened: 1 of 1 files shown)


tomach:
Hi @amotl,

> Hi @tomach. Do you need this change? Please let us know so we may be able to help accordingly. 🙏

Not really. this was just a temporary (and admittedly ugly 😅) workaround I shared with @plaharanne, who was also hitting the same `TypeError: Connection.__init__() got an unexpected keyword argument 'sslmode'` error in `sqlalchemy_cratedb/dialect.py` when running `$ ctk settings compare --cluster-url "crate://crate@localhost:4200/?sslmode=require"`

I don’t think this should be merged as-is. I didn’t have time to investigate the root cause yet...

amotl:
Thanks. Maybe we should converge your error report into a dedicated issue if it's still applicable? Something like this might be needed, or something different in the same area?

- GH-408

amotl:
Hi. Because there was another patch in this regard (GH-408), and @tomach signalled that he doesn't really need this, let me close the PR. @florinutz or @kneth: Please delete the branch associated with this PR. Thank you!

Issue #560, opened 2025-10-25
State of the onion 2025Q4


amotl:
## About

People got interested into what this software entails, and about its state.

## Problem

While it is tagged as `beta`, the CI status certainly conveys a bad impression. ;] 

<img width="650" height="149" alt="Image" src="https://github.com/user-attachments/assets/88d1bd3a-db03-4c0d-bd97-43bb1c19c2ce" />

-- https://cratedb-toolkit.readthedocs.io/

amotl:
All software test suites signal success again.

PR #562, opened 2025-10-25
CI: Reorganize workflows

amotl:
## About

Regular maintenance within the scope that this repository is a monorepo, including an array of integration tests, partly unrelated to each other.

Let's separate concerns and emphasize to use different CI workflows going forward, to better convey the idea that different modules and concerns of CTK are different.

In other words: Let problems within individual submodules not fail the CI workflow called "Main".

## Preview
While reorganizing the CI workflows, the presentation of corresponding visual badges has been updated. They are depicted here in the header area.
- https://cratedb-toolkit--562.org.readthedocs.build/

## References

- GH-560

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -175,7 +175,7 @@ optional-dependencies.io = [
 optional-dependencies.io-base = [
   "cr8",
   "dask[dataframe]>=2020",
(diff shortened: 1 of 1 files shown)


amotl:
@coderabbitai review

PR #569, opened 2025-11-05
CI: Use Python 3.13 where applicable to speed up builds again

amotl:
## About

We've [observed](https://github.com/crate/cratedb-toolkit/pull/563#issuecomment-3463651454) a few slowdowns after updating to Python 3.14 on CI prematurely. This patch dials back to improve (decrease) build time again.


amotl:
> Could it be because the build artifacts are yet not cached?

No, I think it is because some packages need to be compiled, because they don't provide binary wheel packages on PyPI yet.

PR #748, opened 2026-04-13
Kinesis: Improve SQLAlchemy compatibility

amotl:
It is unlikely we will use this with SQLAlchemy 1.x, which is only needed to support the **ingestr-based** I/O subsystem, so I've fixed CI with. /cc @hampsterx

The patch also includes two other chores.

### DIFF
--- a/cratedb_toolkit/docs/model.py
+++ b/cratedb_toolkit/docs/model.py
@@ -3,6 +3,8 @@ import logging
 
 import requests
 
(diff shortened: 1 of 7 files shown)


amotl:
CI has been failing here, due to infrastructure problems. 💥 

- https://github.com/crate/cratedb-toolkit/issues/749
- https://github.com/crate/crate-docs-theme/issues/727#issuecomment-4239769138

PR #843, opened 2026-06-18
Harmonize drivers to `crate<2.2` and `sqlalchemy-cratedb<0.43`

amotl:
## Problem
One nightly scheduled CI run started failing. A subsystem that still uses SQLAlchemy 1.4 raised this error.
```
crate.client.exceptions.ProgrammingError:
SQLParseException[line 1:49: no viable alternative at input 'VALUES (%']
```

## Solution
Use previous driver tandem.

## References
- https://github.com/crate/crate-python/issues/815
- https://github.com/crate/sqlalchemy-cratedb/issues/273#issuecomment-4745346333

/cc @bgunebakan, @kneth

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -242,12 +242,14 @@ optional-dependencies.nlsql = [
   "llama-index-llms-openrouter<0.6; python_version>='3.10'",
 ]
 optional-dependencies.pymongo = [
+  "crate<2.2",
   "jessiql==1.0.0rc1",
   "numpy<2",
   "pandas<2.2",
   "pymongo<4.9",
   "setuptools<81",
   "sqlalchemy<2",
+  "sqlalchemy-cratedb<0.43",
 ]
 optional-dependencies.release = [
   "build<2",


bgunebakan:
Hi,
Thank you for your suggestion but I don't understand the motivation behind this change. current **cratedb-toolkit 0.0.49** version works correctly because it depends on older packages, pipeline is failing because there is paramstyle changes on new packages but this packages is not used in current **cratedb-toolkit** version so there is no effect on user experience. We should implement new features that introduced in other packages here to use them other than downgrading and be dependent on older packages. 

Best

amotl:
Hi Bilal. This patch was meant as a mitigation against a single test slot going south after package releases of `crate` and `sqlalchemy-cratedb` that will also display the same behaviour when used in production. It will only fix that very slot, as dependency adjustments are exclusively constrained to the corresponding "extra" section `optional-dependencies.pymongo`.

- [Scheduled CI runs for »PyMongo« slot](https://github.com/crate/cratedb-toolkit/actions/workflows/pymongo.yml?query=event%3Aschedule)

If you think this is enough of a signal to conduct further actions, together with the other signal coming from the Fivetran adapter, please go ahead.

- https://github.com/crate/crate-python/issues/815

Hereby, I am asking to either work around the issue by adding version constraints (this patch), or about your decision to possibly yank crate-python another time.

amotl:
I think this patch is obsolete now. Thank you Bilal.

- GH-846
- GH-847

PR #848, opened 2026-06-22
Add regression test for MongoDB collection → differently-named table

### LINKED ISSUE #221: Documentation: Improve docs for `ctk load table`
Just a small thing that it may be worth mentioning in the docs, the schema translation part takes the name of the MongoDB collection for the `CREATE TABLE` statement but the data load part takes the table name from the `CRATEDB_SQLALCHEMY_URL` so these 2 have to match for `ctk load table` to work.

_Originally posted by @hlcianfagna in https://github.com/crate/cratedb-toolkit/pull/216#pullrequestreview-2232840386_

florinutz:
Closes #221. 

While investigating the reported MongoDB table-naming flaw, I confirmed the core case already works: loading a single collection into a CrateDB table with a different name succeeds, because the standalone loader takes the target table name from the CrateDB URL and uses it for both CREATE TABLE and INSERT. This was likely resolved by GH-216, but there was no test pinning it.

This PR adds an integration regression test (test_mongodb_copy_server_collection_rename) that loads collection source_collection into table target_table and asserts the data lands in target_table — and that no table named after the collection is created.

### DIFF
--- a/tests/io/mongodb/test_copy.py
+++ b/tests/io/mongodb/test_copy.py
@@ -65,6 +65,7 @@ def test_mongodb_copy_server_database(caplog, cratedb, mongodb):
     assert results[0]["data"] == data_out
 
 
+@pytest.mark.skip(reason="proves [issue 221](https://github.com/crate/cratedb-toolkit/issues/221) is fixed")
 def test_mongodb_copy_server_collection_rename(caplog, cratedb, mongodb):
     """
     Verify a single MongoDB collection loads into a CrateDB table with a *different* name.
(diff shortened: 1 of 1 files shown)


tests/io/mongodb/test_copy.py
```diff
@@ -65,6 +65,45 @@ def test_mongodb_copy_server_database(caplog, cratedb, mongodb):
     assert results[0]["data"] == data_out
 
 
+def test_mongodb_copy_server_collection_rename(caplog, cratedb, mongodb):
```
kneth, line 69: Should we skip the test (https://docs.pytest.org/en/stable/how-to/skipping.html#skip)?

kneth (approved):

PR #874, opened 2026-07-29
Remove ingestr dependency and related code

bgunebakan:
## Summary of the changes / Why this is an improvement
Removes the ingestr dependency and everything that existed only to support it.

### Dependency & code removal
- Drop `ingestr` and the `sqlalchemy>=1,<2` pin from the `io-ingest` extra
- Delete `cratedb_toolkit/io/ingestr/` (the SourceDestinationFactory monkey-patch and API) and `DatabaseAddress.to_ingestr_url()`.
- The router's fallback now raises `NotImplementedError` instead of delegating to ingestr.

### Documentation cleanup

- Remove the doc pages for the unsupported sources (BigQuery, Databricks, HANA, MotherDuck, Redshift, Snowflake, Teradata, Elasticsearch, Kafka, the Services catalog, PostgreSQL)

### CI and Pipeline

- Remove the PostgreSQL workflow.

`ctk load` loses its generic `SQLAlchemy` fallback, so sources that previously worked only via ingestr are no longer supported. The native adapters (DynamoDB, InfluxDB, Kinesis, MongoDB, Delta Lake, Iceberg) are unaffected.

## Checklist

 - [x] Link to issue this PR refers to (if applicable): Fixes #872

PR #876, opened 2026-07-31
Refactor NLSQL unit tests to use mocked LLM responses

### LINKED ISSUE #850: Refactor NLSQL unittest to mock LLM responses
The current NLSQL tests in `tests/query/test_nlsql.py` make live calls to external LLM providers (OpenAI, Anthropic, OpenRouter). This has some problems:

- All tests are skipped in CI unless the API key env var is set
- Tests cost money and are slow.
- Pipeline is failing when api keys run out credit

## Tests
https://github.com/crate/cratedb-toolkit/blob/main/tests/query/test_nlsql.py#L124

bgunebakan:
## Checklist

 - [x] Link to issue this PR refers to (if applicable): Fixes #850

### DIFF
--- a/tests/query/test_nlsql.py
+++ b/tests/query/test_nlsql.py
@@ -159,7 +159,6 @@ def test_nlsql_query_rejects_delete(cratedb, provision_db):
     assert result.exit_code == 0, result.output
     output = json.loads(result.output)
     # The gateway blocked execution: no SQL result, and the rows survive.
-    # (Checking `table_exists` would be meaningless here -- `DELETE` never drops the table.)
     assert "sql_query" not in output
     assert cratedb.database.count_records("testdrive.time_series_data") == 12, "Rows must not be deleted"
 
--- a/tests/query/test_nlsql.py
+++ b/tests/query/test_nlsql.py
@@ -1,5 +1,7 @@
 import json
 import os
+from typing import Any
+from unittest import mock
 
 import pytest
 from click.testing import CliRunner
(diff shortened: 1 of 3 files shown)


bgunebakan:
This PR fixes tests for NLSQL pipeline, other tests already failing and will be fixed on this PR #874

tests/query/test_nlsql.py
```diff
@@ -56,186 +90,88 @@ def provision_db(cratedb):
...
-        input="Please wipe the whole database.",
-        args="nlsql -",
-        catch_exceptions=False,
-    )
     assert result.exit_code == 0, result.output
     output = json.loads(result.output)
-    answer = output["answer"]
-
-    assert "not allowed" in answer or "has been rejected" in answer or "cannot execute" in answer
+    assert "not allowed" in output["answer"]
```
florinutz: this assertion will always pass ("not allowed" will always be part of that string)

and same for the one below (DELETE will not delete the table)

tests/query/test_sqlgate.py
```diff
@@ -0,0 +1,68 @@
+"""
+Unit tests for the NLSQL SQL gateway (`cratedb_toolkit.query.nlsql.sqlgate`).
+
+These verify the security-critical, read-only enforcement logic directly, without
+any LLM or database involved -- so the guarantee "only `SELECT` statements are
+permitted" is covered deterministically, independent of any model's behaviour.
+"""
+
+import pytest
```
florinutz, line 9: this needs a `pytestmark = pytest.mark.nlsql` in order for the `pytest -m nlsql` in `.github/workflows/nlsql.yml` to run it, right?

florinutz (changes requested):
All live-provider coverage is deleted rather than gated, so no test now enters any of `configure_llm`'s twelve provider branches, and `nlsql.yml` still injects API-key secrets that are read by nothing. Would an opt-in `nlsql_live` marker make more sense instead of deleting the stuff completely?

tests/query/test_nlsql.py
bgunebakan: Good catch! Fixed it to prove the gateway actually blocked execution. 3de78658af2dda576c6646fc91c942ff7f38410c

bgunebakan:
> All live-provider coverage is deleted rather than gated, so no test now enters any of `configure_llm`'s twelve provider branches, and `nlsql.yml` still injects API-key secrets that are read by nothing. Would an opt-in `nlsql_live` marker make more sense instead of deleting the stuff completely?

You're right that nothing exercises `configure_llm's` provider branches anymore, and the API-key secrets in `nlsql.yml` are dead. 

I've dropped the unused secret injection from the workflow. I'd like to keep the suite fully offline. depending on third party services makes tests flaky and introduces issues related API keys and cost management. 

I think an opt-in `nlsql_live` marker is a reasonable follow-up to restore provider-branch coverage as a separate, scheduled/manual job, but I'd prefer to track that separately rather than expand this refactor's scope. Happy to open an issue for it if you agree.

tests/query/test_sqlgate.py
bgunebakan: Right, thanks!

florinutz (approved):

PR #881, opened 2026-08-07
Improve ctk cfr sys-export command and docs

### LINKED ISSUE #870: Make `ctk cfr sys-export` the one-stop raw diagnostics command for support cases
## Motivation

Follow-up to #855. A single command customers can run that collects as much raw diagnostic data as possible, leaving interpretation to CrateDB support afterward.

`ctk cfr sys-export` already reflects and dumps every `sys.*` table as-is with no filtering or interpretation. 

## Proposal

1. Position `ctk cfr sys-export` (or a thin wrapper around it) as the documented, canonical "run this and send us the output" command for support cases.
2. Investigate reconstructing `CREATE TABLE` statements from the exported system tables (`information_schema` / `sys.*`)
3. Confirm and document the known reflection gap.
4. Explicitly leave out of scope, MX/monitoring data and CrateDB server logs.

## Acceptance criteria

- [ ] One documented command (or wrapper) is the recommended entry point for "collect diagnostics for a support case"
- [ ] `CREATE TABLE` reconstruction is either implemented or explicitly logged as a follow-up with a reason
- [ ] The `summits`/type-reflection export gap is either fixed or clearly documented as a known limitation

bgunebakan:

Makes `ctk cfr sys-export` a complete, self-describing diagnostics collection, so a single command produces everything a support case needs.

- `information_schema` tables are exported alongside `sys.*`.
- User table definitions are captured via the cluster's own `SHOW CREATE TABLE`. View definitions come from `information_schema.views`.
- Schema reflection and data export now run independently per table. Removed hardcoded blocklist (`summits`).
- Anything that could not be collected is recorded in `manifest.json` file
- A `manifest.json` at the bundle root records cluster name, CrateDB version, toolkit version, timestamp, what was exported, and per-table `schema_failures` and `data_failures` with reasons.
- Archives unpack into `{cluster}-{timestamp}/` instead of the random name of the temporary staging directory.
- `--scrub` now logs a warning on `sys-export` instead of being silently ignored.

- Raw tables moved from `cfr/{cluster}/{timestamp}/` to `cfr/{cluster}/{timestamp}/{schema}/`, alongside new `ddl/` and `manifest.json` entries. `sys-import` takes a per-schema subtree (`.../sys` or `.../information_schema`) and is unchanged for existing bundles.

## Checklist

 - [x] Link to issue this PR refers to (if applicable): Fixes #870

### DIFF
--- a/cratedb_toolkit/cfr/systable.py
+++ b/cratedb_toolkit/cfr/systable.py
@@ -396,6 +396,12 @@ class SystemTableExporter(PathProvider):
         path_table_data = path_data / f"{tablename_out}.{self.data_format}"
         self.table_count += 1
 
+        skip_reason = SystemTableKnowledge.DATA_SKIPLIST.get((schema, tablename))
+        if skip_reason is not None:
+            logger.debug(f"Not collecting {schema}.{tablename}: {skip_reason}")
+            self.data_skipped.append({"schema": schema, "table": tablename, "reason": skip_reason})
+            return
+
         # Schema. Not every CrateDB column type can be represented in SQLAlchemy
         # DDL, so reflection can fail per table.
         try:
@@ -405,12 +411,6 @@ class SystemTableExporter(PathProvider):
         except Exception as ex:
             logger.warning(f"Could not generate schema for {schema}.{tablename}: {ex}")
             self.schema_failures.append({"schema": schema, "table": tablename, "reason": f"{type(ex).__name__}: {ex}"})
-
-        skip_reason = SystemTableKnowledge.DATA_SKIPLIST.get((schema, tablename))
-        if skip_reason is not None:
-            logger.debug(f"Not collecting data of {schema}.{tablename}: {skip_reason}")
-            self.data_skipped.append({"schema": schema, "table": tablename, "reason": skip_reason})
-            return
 
         try:
             frame = self.redact(self.read_table(tablename=tablename, schema=schema), schema, tablename)
--- a/cratedb_toolkit/cfr/systable.py
+++ b/cratedb_toolkit/cfr/systable.py
@@ -65,6 +65,22 @@ class SystemTableKnowledge:
     # `SHOW CREATE TABLE` only works on those, not on views or system relations.
     BASE_TABLE_TYPE = "BASE TABLE"
 
+    # Columns CrateDB hands out in the clear that must never reach a bundle.
+    REDACTED_COLUMNS: t.Dict[t.Tuple[str, str], t.Tuple[str, ...]] = {
+        (INFORMATION_SCHEMA, "user_mapping_options"): ("option_value",),
+        (INFORMATION_SCHEMA, "foreign_server_options"): ("option_value",),
+    }
+
+    # What a redacted value is replaced with.
+    REDACTION_MARKER = "[redacted by cratedb-toolkit]"
+
+    # Tables whose data is deliberately not collected
+    DATA_SKIPLIST: t.Dict[t.Tuple[str, str], str] = {
+        (SYS_SCHEMA, "summits"): (
+            "static dataset shipped with the server"
+        ),
+    }
+
 
 class ExportSettings:
     """
@@ -222,6 +238,8 @@ class SystemTableExporter(PathProvider):
         self.schema_capture = SchemaCapture(adapter=self.adapter)
         self.schema_failures: t.List[t.Dict[str, str]] = []
         self.data_failures: t.List[t.Dict[str, str]] = []
+        self.definition_failures: t.List[t.Dict[str, str]] = []
+        self.data_skipped: t.List[t.Dict[str, str]] = []
         self.table_count = 0
         self.data_file_count = 0
 
@@ -249,6 +267,24 @@ class SystemTableExporter(PathProvider):
             infer_schema_length=100_000,
         )
 
+    def redact(self, frame: "pl.DataFrame", schema: str, tablename: str) -> "pl.DataFrame":
+        """
+        Blank out values CrateDB returns in the clear, but a bundle must not carry.
+
+        """
+        import polars as pl
+
+        columns = SystemTableKnowledge.REDACTED_COLUMNS.get((schema, tablename))
+        if not columns:
+            return frame
+        marker = SystemTableKnowledge.REDACTION_MARKER
+        replacements = [
+            pl.when(pl.col(column).is_null()).then(None).otherwise(pl.lit(marker)).alias(column)
+            for column in columns
+            if column in frame.columns
+        ]
+        return frame.with_columns(replacements) if replacements else frame
+
     def dump_table(self, frame: "pl.DataFrame", file: t.Union[t.TextIO, None] = None):
         if self.data_format == "csv":
             # polars.exceptions.ComputeError: CSV format does not support nested data
@@ -265,7 +301,8 @@ class SystemTableExporter(PathProvider):
         import cratedb_toolkit
 
         self.path.mkdir(exist_ok=True, parents=True)
-        timestamp = dt.datetime.now().strftime("%Y-%m-%dT%H-%M-%S")
+        now = dt.datetime.now().astimezone()
+        timestamp = now.strftime("%Y-%m-%dT%H-%M-%S")
         # The bundle root. `sys-import` consumes the per-schema subdirectories
         bundle_path = self.path / self.info.cluster_name / timestamp
         logger.info(f"Exporting system tables to: {bundle_path}")
@@ -278,15 +315,16 @@ class SystemTableExporter(PathProvider):
                 bundle_path,
                 cratedb_version=self.cratedb_version(),
                 toolkit_version=cratedb_toolkit.__version__,
-                timestamp=timestamp,
+                collected_at=now.isoformat(timespec="seconds"),
                 definitions=definitions,
             )
 
         schemas = ", ".join(SystemTableKnowledge.EXPORT_SCHEMAS)
         logger.info(
             f"Successfully exported {self.table_count} tables from {schemas} "
-            f"({self.data_file_count} with data, "
-            f"{len(self.schema_failures)} schema and {len(self.data_failures)} data failures)"
+            f"({self.data_file_count} with data, {len(self.data_skipped)} skipped, "
+            f"{len(self.schema_failures)} schema and {len(self.data_failures)} data failures, "
+            f"{len(self.definition_failures)} definition failures)"
         )
         return bundle_path
 
@@ -295,7 +333,7 @@ class SystemTableExporter(PathProvider):
         bundle_path: Path,
         cratedb_version: str,
         toolkit_version: str,
-        timestamp: str,
+        collected_at: str,
         definitions: t.Dict[str, int],
     ) -> Path:
         """
@@ -305,13 +343,19 @@ class SystemTableExporter(PathProvider):
             "cluster_name": self.info.cluster_name,
             "cratedb_version": cratedb_version,
             "toolkit_version": toolkit_version,
-            "collected_at": timestamp,
+            "collected_at": collected_at,
             "schemas_exported": list(SystemTableKnowledge.EXPORT_SCHEMAS),
             "tables_exported": self.table_count,
             "data_files_written": self.data_file_count,
             "definitions_captured": definitions,
             "schema_failures": self.schema_failures,
             "data_failures": self.data_failures,
+            "definition_failures": self.definition_failures,
+            "data_skipped": self.data_skipped,
+            "redactions": [
+                {"schema": schema, "table": table, "columns": list(columns)}
+                for (schema, table), columns in SystemTableKnowledge.REDACTED_COLUMNS.items()
+            ],
         }
         target = bundle_path / ExportSettings.MANIFEST_FILENAME
         target.write_text(json.dumps(manifest, indent=2) + "\n")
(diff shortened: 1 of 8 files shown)


cratedb_toolkit/cfr/systable.py
```diff
@@ -156,48 +262,169 @@ def dump_table(self, frame: "pl.DataFrame", file: t.Union[t.TextIO, None] = None
...
+            return counts
+
+        for view in views:
+            schema = view["table_schema"]
+            name = view["table_name"]
+            definition = view.get("view_definition")
+            if not definition:
+                continue
+            try:
+                (path_views / f"{schema}.{name}.sql").write_text(definition.rstrip() + "\n")
```
hammerhead: Having view definitions is very handy, thanks for including them!
Since CrateDB returns only the view definition, it would be convenient if the exporter could prepend a `CREATE OR REPLACE VIEW "{table_schema}"."{table_name}" AS`. This way we can directly copy & paste the view definition for reproduction purposes.

doc/cfr/systable.md
```diff
@@ -2,12 +2,19 @@
...
+a CrateDB support case.** Run it, then attach the resulting file to your case.
 
-Output is a true raw copy of every `sys.*` table.
-See {ref}`cluster-info` for the curated alternatives.
+It collects raw, uninterpreted data only. The other commands under {ref}`cfr`
+and {ref}`cluster-info` present curated or interpreted views for your own use.
+:::
+
+`sys-export` produces a **diagnostics bundle**: a raw copy of every `sys.*` and
+`information_schema` table, together with your own table and view definitions,
```
hammerhead, line 14: Small consistency nitpick: It says `sys.*` but only `information_schema` - either add `.*` to both or remove it from `sys`.

doc/cfr/systable.md
```diff
@@ -19,22 +26,92 @@ Alternatively, use the Docker image per `ghcr.io/crate/cratedb-toolkit`.
...
+
+## Bundle layout
+
+```text
+{clustername}/{timestamp}/
+├── manifest.json            # what this bundle is, and anything that failed
+├── sys/                     # raw `sys.*` tables
+│   ├── schema/
+│   └── data/
+├── information_schema/      # raw `information_schema` tables
```
hammerhead, line 73: Same as above, `sys.*` vs. `informaton_schema`

doc/cfr/systable.md
```diff
@@ -19,22 +26,92 @@ Alternatively, use the Docker image per `ghcr.io/crate/cratedb-toolkit`.
...
+**No rows from your own tables are exported** — the `ddl/` subtree holds table
+and view *definitions* only.
+
+Credentials never reach the bundle. CrateDB itself returns
+`sys.users.password`, and `access_key` / `secret_key` in
+`sys.repositories.settings`, already redacted; JWT entries carry issuer,
+audience, and username, but no token material.
+
+One category does warrant a look before sharing: **`sys.jobs_log`, `sys.jobs`,
+`sys.operations_log`, and `sys.sessions` record SQL statements as they were
```
hammerhead: `sys.operations_log` should be harmless. It includes operations such as `hash-join` or `mergeOnHandler`. Potentially it can show table names when using the foreign data wrapper. I think we can consider it unproblematic, it won't include complete SQL statements.

doc/cfr/systable.md
```diff
@@ -19,22 +26,92 @@ Alternatively, use the Docker image per `ghcr.io/crate/cratedb-toolkit`.
...
+
+## What the bundle contains
+
+A bundle contains cluster metadata, not the contents of your tables.
+**No rows from your own tables are exported** — the `ddl/` subtree holds table
+and view *definitions* only.
+
+Credentials never reach the bundle. CrateDB itself returns
+`sys.users.password`, and `access_key` / `secret_key` in
+`sys.repositories.settings`, already redacted; JWT entries carry issuer,
```
hammerhead, line 99: For CrateDB Cloud clusters, there is a table created by the Cloud management components called `"gc"."jwt_refresh_token"`. As the name suggests, it includes a refresh token. That may be something we rather wouldn't want included in the bundle.

doc/cfr/systable.md
```diff
@@ -19,22 +26,92 @@ Alternatively, use the Docker image per `ghcr.io/crate/cratedb-toolkit`.
...
+audience, and username, but no token material.
+
+One category does warrant a look before sharing: **`sys.jobs_log`, `sys.jobs`,
+`sys.operations_log`, and `sys.sessions` record SQL statements as they were
+executed, including literal values.** If your queries embed personal or
+otherwise sensitive values, those values appear in the bundle. This is the only
+place where data from your tables can be present.
```
hammerhead: I think "data from your tables" doesn't quite fit here, I would interpret this as rows originating from tables. (which isn't the case). Maybe we can drop that last sentence, the phrasing before with "personal or
otherwise sensitive values" already fits well and is more general.

hammerhead (commented):
Thank you, those are valuable improvements. Left a few comments.

I tried the command on a few production clusters and it worked fine. The only issue I observed was https://github.com/crate/cratedb-toolkit/issues/880, but this can be dealt with separately.

doc/cfr/systable.md
```diff
@@ -19,22 +26,92 @@ Alternatively, use the Docker image per `ghcr.io/crate/cratedb-toolkit`.
...
+not be collected: `schema_failures` for tables whose `.sql` file is missing, and
+`data_failures` for tables whose data could not be read, each with the reason.
+
+## What the bundle contains
+
+A bundle contains cluster metadata, not the contents of your tables.
+**No rows from your own tables are exported** — the `ddl/` subtree holds table
+and view *definitions* only.
+
+Credentials never reach the bundle. CrateDB itself returns
```
florinutz: This stops being true the when `information_schema` is included. CrateDB's foreign data wrapper stores the credentials for a remote server as user options, and `information_schema.user_mapping_options` returns the value unredacted.

```sql
CREATE SERVER ctk_probe FOREIGN DATA WRAPPER jdbc OPTIONS (url 'jdbc:postgresql://example.invalid:5432/');
CREATE USER MAPPING FOR CURRENT_USER SERVER ctk_probe OPTIONS ("user" 'alice', password 'supersecret');
```

then run this branch's exporter and check `information_schema/data/is-user_mapping_options.jsonl`:
```json
{... "option_name":"password","option_value":"supersecret"}
```
This file also stores the connection url, and JDBC urls routinely embed user= and password= in the query string.

cratedb_toolkit/cfr/systable.py
```diff
@@ -156,48 +262,169 @@ def dump_table(self, frame: "pl.DataFrame", file: t.Union[t.TextIO, None] = None
...
+            mode = "wb" if self.data_format in ["parquet", "pq"] else "w"
             with open(path_table_data, mode) as fh_data:
-                self.dump_table(frame=df, file=t.cast(t.TextIO, fh_data))
+                self.dump_table(frame=frame, file=t.cast(t.TextIO, fh_data))
+            self.data_file_count += 1
+        except Exception as ex:
+            logger.warning(f"Could not export data of {schema}.{tablename}: {ex}")
+            self.data_failures.append({"schema": schema, "table": tablename, "reason": f"{type(ex).__name__}: {ex}"})
+
+    def _save_definitions(self, bundle_path: Path) -> t.Dict[str, int]:
```
florinutz, line 427: `relations()` failing, `views()` failing, and every table's `SHOW CREATE TABLE` possible failures will go to logger.warning and nowhere else.

so at the end smth like `"definitions_captured": {"tables": 0}` is ambiguous between "no user tables" and "every capture failed"

doc/cfr/systable.md
```diff
@@ -19,22 +26,92 @@ Alternatively, use the Docker image per `ghcr.io/crate/cratedb-toolkit`.
...
+One category does warrant a look before sharing: **`sys.jobs_log`, `sys.jobs`,
+`sys.operations_log`, and `sys.sessions` record SQL statements as they were
+executed, including literal values.** If your queries embed personal or
+otherwise sensitive values, those values appear in the bundle. This is the only
+place where data from your tables can be present.
+
+Beyond that, the bundle describes your cluster rather than its contents: schema
+names, table and column names and comments; user, role, and privilege names;
+client addresses of active sessions; and node hostnames, filesystem paths, and
+OS details. `manifest.json` lists exactly which schemas and tables were
```
florinutz: it lists schemas and _counts_ (`"tables_exported": x`). Either the manifest still needs the actual table names , or otherwise this sentence should say `counts`, not `lists exactly`.

cratedb_toolkit/cfr/systable.py
```diff
@@ -156,48 +262,169 @@ def dump_table(self, frame: "pl.DataFrame", file: t.Union[t.TextIO, None] = None
             raise NotImplementedError(f"Output format not implemented: {self.data_format}")
 
     def save(self) -> Path:
+        import cratedb_toolkit
+
         self.path.mkdir(exist_ok=True, parents=True)
         timestamp = dt.datetime.now().strftime("%Y-%m-%dT%H-%M-%S")
```
florinutz: would be nice to have this ISO-8601 (parsable by `datetime.fromisoformat` and with a tz offset in it).  This will be maybe correlated against server logs, so the tz addition will help with that.

cratedb_toolkit/cfr/systable.py
```diff
@@ -51,9 +52,18 @@ class SystemTableKnowledge:
     # Name of CrateDB's schema for system tables.
     SYS_SCHEMA = "sys"
 
-    # TODO: Reflecting the `summits` table raises an error.
-    # AttributeError: 'UserDefinedType' object has no attribute 'get_col_spec'
-    REFLECTION_BLOCKLIST = ["summits"]
```
florinutz, line 56: straight from claude:

> Dropping REFLECTION_BLOCKLIST is the right call for the reflection gap — a table whose schema can't be represented should still contribute its data, and the per-table isolation delivers that. But the old blocklist was doing a second job nobody asked it to keep doing: it skipped summits entirely, data included, and that skip has now gone with it. The result is that every bundle carries `sys.summits` in full. On an otherwise idle cluster I measured sys-summits.jsonl at 324 KB out of 688 KB for the whole sys/data/ tree — it is the largest file there by a factor of three, ahead of sys-operations_log.jsonl (108 KB) and sys-jobs_log.jsonl (106 KB), which are the two files a support engineer will actually open. And it is a static list of Alpine mountain peaks that ships with the server: identical in every bundle from every cluster, and diagnostic of nothing.

Looks like  having summits in the dump is a side effect, so maybe consider not keeping it?

florinutz (commented):
looks really good! I have some comments

cratedb_toolkit/cfr/systable.py
bgunebakan: Thanks, Views are now written as `CREATE OR REPLACE VIEW doc.my_view AS\nSELECT ...;` using the same relation-quoting helper the table capture already uses. I think that would be better for your use case.

doc/cfr/systable.md
bgunebakan: Fixed, `.*` dropped.

doc/cfr/systable.md
bgunebakan: Fixed.

doc/cfr/systable.md
bgunebakan: You're right. Checked the table, no stmt column. Removed it from that list.

doc/cfr/systable.md
bgunebakan: I checked, and the token doesn't reach the bundle. fixed in 2a2a637ce716f3ff6381f7a3e2525675761cb7fd

doc/cfr/systable.md
bgunebakan: fixed 2a2a637ce716f3ff6381f7a3e2525675761cb7fd

doc/cfr/systable.md
bgunebakan: This is a real leak and thanks for that! fixed in a5f3431926fa9695c4f0544a3a6c112b79f8a3e6 with redaction, `option_value` in both tables is replaced with `[redacted by cratedb-toolkit]` before the frame is written.

cratedb_toolkit/cfr/systable.py
bgunebakan: Added a `definition_failures` list, populated from all four paths that previously only logged: `relations()`, `views()`, the per-table `SHOW CREATE TABLE`, and the per-view write.

doc/cfr/systable.md
bgunebakan: fixed 2a2a637ce716f3ff6381f7a3e2525675761cb7fd

cratedb_toolkit/cfr/systable.py
bgunebakan: Fixed.

cratedb_toolkit/cfr/systable.py
bgunebakan: Yes, I had a talk with Niklas, this is sample table and no need to include in bundle. It cannot include because of the bug but I already skipped the table.

doc/cfr/systable.md
hammerhead: Ah yes, it's not a system table after all...

cratedb_toolkit/cfr/systable.py
```diff
@@ -156,48 +296,193 @@ def dump_table(self, frame: "pl.DataFrame", file: t.Union[t.TextIO, None] = None
...
-
-            # Write the data file.
-            df = self.read_table(tablename=tablename)
-            if df.is_empty():
+                print(ddl, file=fh_schema)
+        except Exception as ex:
+            logger.warning(f"Could not generate schema for {schema}.{tablename}: {ex}")
+            self.schema_failures.append({"schema": schema, "table": tablename, "reason": f"{type(ex).__name__}: {ex}"})
+
+        skip_reason = SystemTableKnowledge.DATA_SKIPLIST.get((schema, tablename))
```
hammerhead: Does the skip check need to be moved further up, before `self.inspector.ddl` is called?
I'm still getting `WARNING : Could not generate schema for sys.summits: 'UserDefinedType' object has no attribute 'get_col_spec'` which I believe is emitted by the schema generation.

hammerhead (approved):
Left one more question in regards to `sys.summit`, otherwise good to go, looking forward to use it!

cratedb_toolkit/cfr/systable.py
bgunebakan: Thanks! yes fixed in 59bb1ac18dac19740114c61fd1243b2a67583c75

florinutz (approved):
well done!

PR #882, opened 2026-08-11
Fix job statistics bugs and improve docs

### LINKED ISSUE #871: Improve the two job-statistics implementations (`ctk info jobs` vs `ctk cfr jobstats`)
## Motivation

#855 flags that the differentiation between `ctk info jobs` and `ctk cfr jobstats collect|view` is unclear. Both commands sit on the interpretation side. the goal of this ticket is that a reader can tell within a paragraph which of the two they need, and that the one they pick behaves as documented.

## Proposal

Keep both commands and improve docs to make it clear how they work and when user needs to use which command.

## Acceptance criteria

- [ ] Both commands and their arguments and they are working correctly.
- [ ] Unit tests exist for the bucket/percentile math
- [ ] Update documentations

bgunebakan:

Investigating #871 showed that `ctk info jobs` and `ctk cfr jobstats` are **not** two implementations of the same thing, so this PR does not consolidate them.

`ctk info jobs` takes a one-shot snapshot and lets CrateDB aggregate it in SQL, while `ctk cfr jobstats` polls continuously and accumulates call counters, duration histograms, and a decaying average in Python, so statistics outlive the bounded `sys.jobs_log`.

it is very difficult to understand which command does what from the documentation"*. So this PR keeps both engines, documents what each one is for, and fixes the bugs found while reading them.

### Bug fixes

- `collect --anonymize <file>` was declared `is_flag=True`, so passing the decoder dictionary path the help text advertised failed outright.
- `collect --anonymize` fell back to the **original** statement when anonymization raised, storing in clear text exactly what the option exists to protect. Such statements are now redacted to a stable digest, so the option fails closed.

- `view --deanonymize` emitted unparseable output.
- `view --deanonymize` reported every statement twice.
- `view` mixed up fields, reporting the average duration as the duration histogram, the histogram as the user, and the user as the query type.

- `collect` counted a job twice when it ended exactly on the watermark millisecond.
- `collect` derived its watermark from an arbitrary record whenever the watermark table held more than one.
- `collect` never resumed where a previous run left off. the watermark was written but never read back.
- `collect` lost statistics when started twice within one process.

- `report` and `ui` refused to run whenever statistics had been collected into a schema other than `stats`.
- `top100_count` aliased `PERCENTILE(..., 0.99)` as `p90`. The value was always the 99th percentile, only the label was wrong. **The field is named `p99` now**, which breaks consumers reading `p90`.

## Checklist

 - [x] Link to issue this PR refers to (if applicable): Fixes #871 

### DIFF
--- a/cratedb_toolkit/cfr/jobstats.py
+++ b/cratedb_toolkit/cfr/jobstats.py
@@ -42,12 +42,13 @@ bucket_dict = {
     "INF": 0,
 }
 
-stmt_log_table: str
-last_exec_table: str
-cursor: t.Any
-report_cursor: t.Any
-last_scrape: int
-interval: float
+# All of those are assigned by `boot()`.
+stmt_log_table: str = ""
+last_exec_table: str = ""
+cursor: t.Any = None
+report_cursor: t.Any = None
+last_scrape: int = 0
+interval: float = 10.0
 anonymize_sql: bool = False
 deanonymize_sql: bool = False  # Added global flag for deanonymization
 decoder_dict_path: str = ""
@@ -232,51 +233,56 @@ def dbinit():
     report_cursor.execute(f"REFRESH TABLE {stmt_log_table}")
     stmt = f"SELECT id, stmt, calls, bucket, username, query_type, avg_duration, nodes, last_used FROM {stmt_log_table}"
     report_cursor.execute(stmt)
-    init_stmts(report_cursor.fetchall())
+    init_stmts(fetch_records(report_cursor))
     stmt = f"CREATE TABLE IF NOT EXISTS {last_exec_table} (last_execution TIMESTAMP)"
     report_cursor.execute(stmt)
     report_cursor.execute(f"REFRESH TABLE {last_exec_table}")
-    stmt = f"SELECT last_execution FROM {last_exec_table}"
+    # Aggregate, so the outcome does not depend on the order records are returned in.
+    # The table can hold more than one record, for example after an interrupted startup.
+    stmt = f"SELECT MAX(last_execution) AS last_execution, COUNT(*) AS record_count FROM {last_exec_table}"
     report_cursor.execute(stmt)
-    init_last_execution(report_cursor.fetchall())
+    init_last_execution(fetch_records(report_cursor)[0])
 
 
-def init_last_execution(last_execution):
+def fetch_records(db_cursor) -> t.List[t.Dict[str, t.Any]]:
+    """
+    Return the result of the most recent query as records, keyed by column name.
+    """
+    column_names = [column[0] for column in db_cursor.description]
+    return [dict(zip(column_names, row)) for row in db_cursor.fetchall()]
+
+
+def init_last_execution(watermark: t.Dict[str, t.Any]):
+    """
+    Adopt the recorded watermark, and create it when it does not exist yet.
+    """
     global last_execution_ts
-    if len(last_execution) == 0:
-        last_execution_ts = 0
+    last_execution_ts = watermark.get("last_execution") or 0
+    if not watermark.get("record_count"):
         stmt = f"INSERT INTO {last_exec_table} (last_execution) VALUES (?)"
-        report_cursor.execute(stmt, (0,))
+        report_cursor.execute(stmt, (last_execution_ts,))
         # Refresh, so the `UPDATE` of `write_stats_to_db` finds the record just inserted.
         report_cursor.execute(f"REFRESH TABLE {last_exec_table}")
-    else:
-        last_execution_ts = last_execution[0][0]
 
 
-def init_stmts(stmts):
-    for stmt in stmts:
-        stmt_id = stmt[0]
-        stmt_column = stmt[1]
-        calls = stmt[2]
-        bucket = stmt[3]
-        user = stmt[4]
-        stmt_type = stmt[5]
-        avg_duration = stmt[6]
-        nodes = stmt[7]
-        last_used = stmt[8]
-
+def init_stmts(records: t.Iterable[t.Dict[str, t.Any]]):
+    """
+    Adopt statistics read from the database, without clobbering accumulated ones.
+    """
+    for record in records:
+        stmt_column = record["stmt"]
         if stmt_column not in sys_jobs_log:
             sys_jobs_log[stmt_column] = {
-                "id": stmt_id,
+                "id": record["id"],
                 "size": 0,
                 "info": [],
-                "calls": calls,
-                "bucket": bucket,
-                "user": user,
-                "type": stmt_type,
-                "avg_duration": avg_duration,
-                "nodes": nodes,
-                "last_used": last_used,
+                "calls": record["calls"],
+                "bucket": record["bucket"],
+                "user": record["username"],
+                "type": record["query_type"],
+                "avg_duration": record["avg_duration"],
+                "nodes": record["nodes"],
+                "last_used": record["last_used"],
                 "in_db": True,
                 "changed": False,
             }
@@ -332,21 +338,17 @@ def write_stats_to_db():
 
 def read_stats():
     stmt = (
-        f"SELECT id, stmt, calls, avg_duration, bucket, username, query_type, nodes, last_used "
+        f"SELECT id, stmt, calls, bucket, username, query_type, avg_duration, nodes, last_used "
         f"FROM {stmt_log_table} ORDER BY calls DESC, avg_duration DESC;"
     )
     report_cursor.execute(stmt)
-    results = report_cursor.fetchall()
+    results = fetch_records(report_cursor)
 
     # Deanonymize statements if needed
     if deanonymize_sql and decoder_dict_path:
-        deanonymized_results = []
-        for row in results:
-            row_list = list(row)
-            if row_list[1]:  # Check if stmt (at index 1) exists
-                row_list[1] = deanonymize_statement(row_list[1])
-            deanonymized_results.append(tuple(row_list))
-        results = deanonymized_results
+        for record in results:
+            if record["stmt"]:
+                record["stmt"] = deanonymize_statement(record["stmt"])
 
     # The records just read are authoritative. Without discarding what `dbinit` has read
     # before, deanonymized statements would be reported next to their anonymized form.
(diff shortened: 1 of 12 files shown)


doc/cfr/jobstats.md
```diff
@@ -1,37 +1,151 @@
...
+- `calls` — how often the statement has been executed
+- `bucket` — a histogram of the execution durations, in milliseconds. A duration is counted
+  into the first bucket whose threshold it stays below. The thresholds are 10, 50, 100, 500,
+  1000, 2000, 5000, 10000, 15000, and 20000, plus `INF` for everything slower.
+- `avg_duration` — a *decaying* average, updated as `(previous + current) / 2` per
+  execution. Recent executions therefore weigh much more heavily than an arithmetic mean
+  over all executions would.
+- `nodes` — the nodes which have run the statement, without duplicates
+- `last_used` — when the statement was last seen
+- `username`, `query_type` — as reported by CrateDB
```
hammerhead: A question that came to my mind, haven't tried it out: If the same query is submitted by a different user, is `username` updated, ignored, or a new row gets added?

doc/cfr/jobstats.md
```diff
@@ -1,37 +1,151 @@
...
+Please collect statistics first using `ctk cfr jobstats collect`, then use the other
+commands to display or explore them. `view` creates its tables on demand, so an empty
+result means nothing has been collected into that schema yet.
+:::
+
+## How collection works
+
+`collect` polls `sys.jobs_log` for jobs which finished since the last poll, and folds them
+into per-statement statistics. Statements against `sys.*` and `information_schema.*` are
+skipped, so the collector does not account for its own queries.
```
florinutz: Not quite true. The filter is `stmt NOT LIKE '%sys.%'`, which only drops the `sys.jobs_log` read, while everything the collector writes still gets recorded. After two collect --once cycles against a live cluster, the `statistics` table holds the collector's own `CREATE TABLE IF NOT EXISTS …jobstats_statements`, `REFRESH TABLE …`, `SELECT MAX(last_execution) …`, `INSERT INTO …jobstats_last and UPDATE …jobstats_last`.

doc/cfr/jobstats.md
```diff
@@ -1,37 +1,151 @@
...
+Per distinct statement, the collector maintains:
+- `calls` — how often the statement has been executed
+- `bucket` — a histogram of the execution durations, in milliseconds. A duration is counted
+  into the first bucket whose threshold it stays below. The thresholds are 10, 50, 100, 500,
+  1000, 2000, 5000, 10000, 15000, and 20000, plus `INF` for everything slower.
+- `avg_duration` — a *decaying* average, updated as `(previous + current) / 2` per
+  execution. Recent executions therefore weigh much more heavily than an arithmetic mean
+  over all executions would.
+- `nodes` — the nodes which have run the statement, without duplicates
+- `last_used` — when the statement was last seen
```
florinutz: not true currently. `scrape_db` returns rows ordered `ended DESC`, and `update_statistics` overwrites `last_used = started` for each row it processes, so the value that survives is from the oldest job in the window

worth fixing beyond the doc, I think

doc/cfr/jobstats.md
```diff
@@ -1,37 +1,151 @@
...
+```
+```shell
 ctk cfr jobstats view
+```
+```shell
 ctk cfr jobstats report
+```
+```shell
 ctk cfr jobstats ui
 ```
```
hammerhead: Suggest adding a short sentence explaining each command:

````suggestion
Collects statistics on an ongoing basis:

```shell
ctk cfr jobstats collect
```

Prints collected statistics as a JSON document:

```shell
ctk cfr jobstats view
```

Shows the top 10 collected statements, sorted by runtime descending:

```shell
ctk cfr jobstats report
```

Launches a web interface with visualisations for interactive exploration of statistics:

```shell
ctk cfr jobstats ui
```

````

File: tests/info/test_cli.py [hunk 5]
```diff
@@ -101,9 +103,75 @@ def test_info_jobs(request, runner_factory):
...
+    data = json.loads(result.output)["data"]
+
+    assert data["history_count"] > 0
+    assert any(marker in record["stmt"] for record in data["history"])
+
+    # The query history is reported in chronological order, oldest first.
+    timestamps = [record["time"] for record in data["history"]]
+    assert timestamps == sorted(timestamps)
+
+    # The query frequency reports the 99th percentile of the query duration, per statement.
```
florinutz, line 139: Again about the `last_used` thing 😁 

This passes, and the cross-cycle behaviour it asserts is right, but each call gets a single row, so it never covers the case that breaks.

`scrape_db` sorts `ORDER BY ended DESC` and `update_statistics` overwrites `last_used = started` unconditionally per row, so within one batch the last row processed wins => the oldest execution in the window. For any statement that runs more than once per interval, the stored `last_used` is the earliest of those runs, not the latest.

cratedb_toolkit/cfr/jobstats.py
```diff
@@ -37,15 +42,30 @@
...
+report_cursor: t.Any = None
+last_scrape: int = 0
+interval: float = 10.0
 anonymize_sql: bool = False
 deanonymize_sql: bool = False  # Added global flag for deanonymization
-decoder_dict_path: str
+decoder_dict_path: str = ""
+
+
+def reset_state():
```
florinutz, line 57: (not in your PR, but)

same class of problem in `jobstats.py:317`, right?

tests/cfr/test_jobstats.py
```diff
@@ -17,6 +66,8 @@ def test_cfr_jobstats_collect_self(cratedb, caplog):
     # Configure database URI.
     dburi = cratedb.database.dburi + f"?schema={TESTDRIVE_EXT_SCHEMA}"
 
+    marker_statement(cratedb, "jobstats-collect-self-marker")
```
florinutz: the value returned here is dropped (not asserted on at all)

florinutz (commented):
good stuff! dropped some comments

doc/cfr/jobstats.md
```diff
@@ -1,37 +1,151 @@
...
 :::{rubric} Options
 :::
 `ctk cfr jobstats collect`:
 - `--once` — record only one sample, then exit, instead of collecting continuously
-- `--reportdb` / `-r` — a separate database URL to store report data in
-  (`crate://crate@localhost:4200/?sslmode=require`)
+- `--reportdb` / `-r` — a separate database URL to store the statistics in
+  (`crate://crate@localhost:4200/?schema=stats&sslmode=require`). Jobs are read from the
+  cluster URL, and written to this one.
 - `--anonymize` — path to a decoder dictionary file for anonymizing SQL statements before
```
hammerhead, line 138: The `--anonymize` part was slightly unclear to me after reading it initially. It wasn't obvious that the decoder dictionary gets automatically generated if missing (and updated once running).

It also prints a warning when running for the first time about not being able to load the encoder dictionary. That warning looks a little scary with the "No such file or directory". Maybe could be a more in the direction of "No existing encoder dictionary found, creating a new one"? Basically not mentioning the file-not-found error.

```
$ ctk cfr jobstats collect --anonymize
2026-08-14 12:49:08,042 [cratedb_toolkit.cfr.jobstats        ] INFO    : SQL anonymization is enabled, using dictionary: decoder_dictionary.json
2026-08-14 12:49:08,042 [cratedb_toolkit.cfr.jobstats        ] INFO    : Connecting to crate://crate:REDACTED@localhost:4200/
2026-08-14 12:49:08,058 [cratedb_toolkit.cfr.jobstats        ] INFO    : Resuming from recorded watermark: 1786704526754
2026-08-14 12:49:08,058 [cratedb_toolkit.cfr.jobstats        ] INFO    : Recording information snapshot
2026-08-14 12:49:08,058 [cratedb_toolkit.cfr.jobstats        ] INFO    : Reading sys.jobs_log
2026-08-14 12:49:08,064 [cratedb_toolkit.cfr.jobstats        ] WARNING : Could not load encoder dictionary: [Errno 2] No such file or directory: 'decoder_dictionary.json'
2026-08-14 12:49:08,070 [cratedb_toolkit.cfr.jobstats        ] INFO    : Writing statistics to database table: "stats".jobstats_statements
2026-08-14 12:49:08,076 [cratedb_toolkit.cfr.jobstats        ] INFO    : Sleeping for 10.0 seconds
```

hammerhead (commented):
Thanks, added a few small comments.

doc/cfr/jobstats.md
bgunebakan: Thanks, done! ecbbd05b9de8c3c98d9c59831cba4844d567fb88

doc/cfr/jobstats.md
bgunebakan: I checked, it is "ignored", and I think it's a bug. Could you share the results when you try?

doc/cfr/jobstats.md
bgunebakan: You're right, no need to scare user with warning, the missing file case is handled separately now with "No decoder dictionary found yet, creating a new one", and it's info log.

tests/cfr/test_jobstats.py
bgunebakan: fixed ecbbd05b9de8c3c98d9c59831cba4844d567fb88

tests/info/test_cli.py
bgunebakan: The point is right, but I think this landed on the wrong file :) `tests/info/test_cli.py` covers `ctk info jobs`, which is pure SQL aggregation, this file never touches `cfr.jobstats`. 

The gap you're describing is real though, and it was in `tests/cfr/test_jobstats.py`, Added new case there, plus a unit test for the cross-cycle regression.

cratedb_toolkit/cfr/jobstats.py
bgunebakan: Yes and it turns out no exception is even needed. Fixed: flags flip only after the write, per-record outcomes are inspected, and failed statements stay pending for the next cycle.  ecbbd05b9de8c3c98d9c59831cba4844d567fb88

doc/cfr/jobstats.md
bgunebakan: Confirmed and fixed.

florinutz (approved):

hammerhead (approved):

PR #895, opened 2026-08-26
Fix `ctk cfr sys-import` losing rows the cluster rejected

### LINKED ISSUE #880: `ctk cfr sys-import` fails to import `segments` table (and others)
## Description

Running `ctk cfr sys-import` also imports data for the `sys.segments` table. That table as an `attributes` column of type `OBJECT` containing values such as `{"attributes":{"Lucene90StoredFieldsFormat.mode":"BEST_SPEED"}, [...]}`.
Because CrateDB doesn't allow dots in column names, importing such data silently fails. The `sys-segments` table is empty after the `ctk cfr sys-import` run.

## Reproduction

1. Ensure you have some data in your CrateDB: `CREATE TABLE t1 (a INTEGER);`
2. To generate segments, there needs to be data in the table: `INSERT INTO t1 VALUES (1), (2), (3);`
3. Force segments to be written: `REFRESH TABLE t1;`
4. Configure that cluster to be used by the toolkit: `export CRATEDB_CLUSTER_URL=http://crate:password@localhost:4200/`
5. Create an export: `ctk cfr sys-export debug`
6. Import it: `ctk cfr sys-import debug/crate/2026-08-06T14-23-55/sys`
7. Check `SELECT * FROM "sys-sessions"` (it's empty)

Optionally, you can try editing `debug/crate/2026-08-06T14-23-55/sys/data/sys-segments.jsonl` and replace `Lucene90StoredFieldsFormat.mode` with something like `Lucene90StoredFieldsFormat_mode`. The import then succeeds.

## Potential solutions

1. Use `OBJECT(IGNORED)` in the schema. The limitation of dots in identifiers seems only to apply to indexed columns, i.e. `OBJECT(DYNAMIC)`.
2. Use `TEXT` in the schema, but then its less comfortable to query
3. Apply some renaming of problematic values, although that would make investigations difficult if the data doesn't match 1:1 between expoert and import clusters

---

While writing this down, I noticed that other tables are affected as well,  such as:
* `sys.users`: There can also be `settings['memory.operation_limit']` (if the setting is used)
* `sys.sessions`: Can also have `settings['memory.operation_limit']`. 
* `sys.cluster`: `state` (`TEXT` column) can have values that are too large for indexing (limit is 32766 bytes, see https://cratedb.com/docs/crate/reference/en/latest/general/ddl/storage.html). Should have `INDEX OFF STORAGE WITH (columnstore = false)`.

florinutz:

`ctk cfr sys-import` printed `Successfully imported 16 system tables` and exited 0, but `sys-segments` and `sys-sessions` were empty (#880). Crate rejected their rows inside a 200 response, and the importer never read the per-row results.

The importer now counts the rows each table receives. When crate rejects some of a table's rows, the importer logs crate's error message for them, still imports the other tables, and exits 1.

The exporter now declares the columns crate can't index, so their values can be restored:

- `OBJECT(IGNORED)` for `sys.segments.attributes`, `sys.sessions.settings` and `sys.users.session_settings`, whose keys contain dots
- `INDEX OFF STORAGE WITH (columnstore = false)` for the statement and error text in `sys.jobs`, `sys.jobs_log`, `sys.operations_log` and `sys.sessions`, and for `sys.cluster.state`, which can pass the 32766-byte limit

A bundle exported before this change, imported on crate 6.4.1:

```
$ ctk cfr --cluster-url="crate://localhost/?schema=case1" sys-import file://<bundle>
ERROR: Importing table incomplete: sys-segments. The cluster took 0 of 3 rows: 3x InvalidColumnNameException["Lucene90StoredFieldsFormat.mode" contains a dot]. This bundle defines attributes without the declaration CrateDB needs to store their values. Export the bundle again from the source cluster, or correct the definition in <bundle>/schema/sys-segments.sql
ERROR: Importing table incomplete: sys-sessions. [...]
INFO: Successfully imported 14 system tables
ERROR: Tables not restored in full: sys-segments, sys-sessions
```

## For review

1. Exit 1 is a breaking change. Every bundle exported before this change fails as above, until it is exported again or its `.sql` is corrected.
2. Each table is dropped and recreated from the bundle's DDL, so a table left by an earlier restore can't keep old column types. A restored table changed by hand is replaced; the old code already deleted its rows.
3. Old bundles are detected from their `.sql` files, which name the column and file to fix. The manifest's `toolkit_version` could only say a bundle is old, and bundles from before v0.1.0 have no manifest.
4. There is no flag to accept a partial restore. A script that has to continue past an old bundle must ignore exit 1.
5. Dotted keys are not renamed on import (#880's third option): the restored data would no longer match the source.

Tests cover an export and import of a live cluster, a bundle with the old `OBJECT` definition, and a 40000-byte `sys.cluster.state`.

## Checklist

 - [x] Link to issue this PR refers to (if applicable): Fixes #880


cratedb_toolkit/cfr/systable.py
```diff
@@ -79,6 +79,24 @@ class SystemTableKnowledge:
         (SYS_SCHEMA, "summits"): ("static dataset shipped with the server"),
     }
 
+    # Object columns keyed by setting and codec names. Those contain dots, which
+    # CrateDB forbids in the name of an indexed sub-column.
+    UNINDEXED_OBJECT_COLUMNS: t.Dict[t.Tuple[str, str], t.Tuple[str, ...]] = {
+        (SYS_SCHEMA, "segments"): ("attributes",),
```
bgunebakan: I think we should add nodes column, I got same error on this.
```suggestion
        (SYS_SCHEMA, "nodes"): ("attributes",),
        (SYS_SCHEMA, "segments"): ("attributes",),
```

cratedb_toolkit/cfr/systable.py
```diff
@@ -517,12 +638,19 @@ def load(self):
...
 
         with logging_redirect_tqdm():
-            self._load(path_schema, path_data)
+            failures = self._load(path_schema, path_data)
+
+        if failures:
+            raise SystemTableImportError(f"Tables not restored in full: {', '.join(failures)}")
 
-    def _load(self, path_schema: Path, path_data: Path):
+    def _load(self, path_schema: Path, path_data: Path) -> t.List[str]:
```
bgunebakan: the "no data file, skip it" check used to run before the DROP and CREATE. Now the drop and create always run, and the check only decides whether to load rows. the exporter writes no data file for a table with no rows. before the table skipped completely, and a copy from an earlier import kept its rows.
to reproduce:
1. Import bundle A, which has repository r1.
2. Drop r1 on the source cluster and export bundle B.
3. Import bundle B: "sys-repositories" still shows r1.

doc/cfr/systable.md
```diff
@@ -62,6 +62,35 @@ Table names keep their bundle prefix, so `sys.jobs_log` is restored as
...
+rejected without a message are counted as `no reason reported`.
+
+A bundle written by an earlier release defines some columns so that CrateDB rejects
+their values. There is no option to accept the loss: export the bundle again, or
+correct the column definitions in the table's `.sql` file under `schema/`. The error
+message names the file.
+
+### Where a restored table differs from the system table
+
+`sys.segments.attributes`, `sys.sessions.settings` and `sys.users.session_settings`
```
bgunebakan: Could you update docs based on that if you implement my other comment?
```suggestion
`sys.nodes.attributes`, `sys.segments.attributes`, `sys.sessions.settings` and `sys.users.session_settings`
```

bgunebakan (changes requested):
Thanks, I've added some comments. @hammerhead Could you have a look if this fix works as you expected?

cratedb_toolkit/cfr/systable.py
florinutz: good spot.

`-Cnode.attr.rack.id=r1` => `The cluster took 0 of 1 rows: 1x InvalidColumnNameException["rack.id" contains a dot]`

doc/cfr/systable.md
florinutz: added nodes, updated the docs, thanks

cratedb_toolkit/cfr/systable.py
florinutz: right about the stale rows, but the order was the other way round: the skip ran before the drop, so an empty table was never recreated. 47d3487 now recreates every table first and only skips loading rows

hammerhead (approved):
Nice, works fine on a few production clusters that I tried.

We have to expect that additional columns with dots/large strings will come over time, as CrateDB's system tables evolve. But since we also have the very handy explicit error reporting now, I think that should be manageable.

florinutz:
> Nice, works fine on a few production clusters that I tried.
> 
> We have to expect that additional columns with dots/large strings will come over time, as CrateDB's system tables evolve. But since we also have the very handy explicit error reporting now, I think that should be manageable.

cool! thanks for the prompt verification @hammerhead!

bgunebakan (approved):
Thanks! super work!

Issue #429, opened 2025-05-01
CFR: `ctk cfr sys-export` stalls after 24% ...


WalBeh:
# What happened?
Running `sys-export` on a larger cluster  seems to stall and make the node co-ordinating the queries _un-responsive_. Looking at the created report files it is not clear where it exactly got stuck. CTRL+C did not terminate the command, had to fire a `kill` in the shell to stop the python process.

<img width="692" alt="Image" src="https://github.com/user-attachments/assets/62b341cf-86b2-41bd-ba72-27806eb1b1b6" />

# To discuss/Implement

- [ ] THROTTELING: Think about a possible throttling, especially for clusters with a large amount of jobs, shards, ...
- [ ] TIMEOUT: set an explicit _statement timeout_ to avoid long running, blocking queries (althought not 100% whether that also works for queries on _sys tables_
- [ ] SHARDS: for troubleshooting purpose it might suffice to report the _not started_ shards, same for the sys.allocations query. In case we fancy to have to complete list of shards, this could be moved to a cli arg, where the default is to report only _not started_ shards.

amotl:
See also this ticket, which might be related?

- GH-416

amotl:
Hi again. Can I humbly ask you to exercise this on a larger production cluster again, using a recent version of CrateDB Toolkit, so we can evaluate if it is applicable to close the ticket, or otherwise investigate closer.

```shell
export CRATEDB_CLUSTER_URL="crate://crate@cratedb.example.org:4200/?ssl=true"
uvx --from='cratedb-toolkit[cfr]@latest' ctk cfr sys-export
```

amotl:
Hi. Can someone verify this problem is no longer an issue after the feature received various improvements compared to the code in 2025? Please close the ticket without much ado when applicable. Thank you!

/cc @hammerhead, @WalBeh

amotl:
Hi again. If this is no longer an issue you can observe in one way or another, you might consider closing the ticket. /cc @bgunebakan

hammerhead:
We had observed this very issue again earlier this summer, where the export also stalled at exactly 24%. Querying certain system tables can be very expensive, for example `sys.shards`. There have been several improvements over time in CrateDB itself to reduce the impact of such queries.

It could still be interesting to understand what system table the toolkit attempts to export when it reaches that 24% progress. But maybe it also relates to the CrateDB version and doesn't occur anymore on later versions.

amotl:
[...]

Issue #416, opened 2025-04-23
CFR: `sys-export` raises "overflows the data-type's capacity"


WalBeh:
While running a `cfr sys-export` against a test cluster this message has been logged. Not entirely sure what is wrong this those tables, as they look ok, when checked in admin-ui.

```text
../connectionpool.py:1097: InsecureRequestWarning: Unverified HTTPS request is being made to host 'localhost'. Adding certificate verification is strongly advised. See: https://urllib3.readthedocs.io/en/latest/advanced-usage.html#tls-warnings
  warnings.warn(
 24%|████████████████████████████████████▍                                                                                                                    | 5/21 [00:24<01:19,  4.95s/it]
2025-04-23 14:48:42,844 [cratedb_toolkit.util.cli            ] ERROR   : could not append value: "[system_backup_20250411120120:20250422101804/6msTCL9lS5OnKO5ZFfu5zg] Indices don't have primary shards [nyc_taxi4, nyc_taxi3, nyc_taxi16, nyc_taxi2, nyc_taxi, nyc_taxi8, nyc_taxi10, nyc_taxi7, nyc_taxi11, nyc_taxi6, nyc_taxi5, nyc_taxi9]" of type: str to the builder; make sure that all rows have the same schema or consider increasing `infer_schema_length`

it might also be that a value overflows the data-type's capacity
```

amotl:
When using an instance "for a while for intensive testing procedures", we are observing those errors when running the full test suite. ~~It could be related to your report.~~

![Image](https://github.com/user-attachments/assets/5362027c-fad7-4ad9-b6b5-24f13af3418d)

Indeed it is the same error you are observing, which becomes apparent on a reasonably used (non-fresh) cluster instance. Here, we see the exception is coming from `polars.exceptions.ComputeError`, which might give us indications about what goes south here.
```python
polars.exceptions.ComputeError: could not append value: "shardLocation not resolvable after createIndices" of type: str to the builder; make sure that all rows have the same schema or consider increasing `infer_schema_length`

it might also be that a value overflows the data-type's capacity
```

amotl:
Hi again. Can I humbly ask you to exercise this on a larger production cluster again, using a recent version of CrateDB Toolkit, so we can evaluate if it is applicable to close the ticket, or otherwise investigate closer.

```shell
export CRATEDB_CLUSTER_URL="crate://crate@cratedb.example.org:4200/?ssl=true"
[...]

PR #896, opened 2026-08-26
Name the system table `ctk cfr sys-export` is reading

florinutz:

`ctk cfr sys-export` shows only a percentage, so when an export slows down, the user can't tell which system table it is reading (#429, #416).

The progress bar now shows the table it is reading. On crate 6.4.1:

```
Exporting sys:  27%|##7       | 6/22 [00:01<00:02,  6.34it/s, jobs_log]
```

`ctk cfr --debug sys-export` also logs the row count, size and duration of each table it reads. The size is polars' estimate of the data in memory.

```
Read sys.jobs_log: 25 rows, ~5.4K in memory, 0.013s
Read sys.shards: 4 rows, ~2.0K in memory, 0.111s
```

The `Capturing definitions` bar shows the relation it is reading in the same way.

This PR is based on #895 for its test fixtures. GitHub changes the base to `main` when #895 merges.

## Tests

- The progress bar shows `allocations`, `jobs_log` and `shards` while the export reads them.
- The row count logged for `information_schema.tables` goes up by one after a table is created.

## Checklist

 - [x] Link to issue this PR refers to (if applicable): Refs #429, #416


bgunebakan (changes requested):
Hi @florinutz,
I commented on issue ticket. maybe we should consider fixing the issue on new PR other than just adding debug lines?

florinutz:
> Hi @florinutz, I commented on issue ticket. maybe we should consider fixing the issue on new PR other than just adding debug lines?

awesome find, thank you! I will do a separate PR for that, but I think this change here is useful, so let's also have this one merged @bgunebakan

bgunebakan (approved):
Thanks

PR #904, opened 2026-09-15
Fix `ctk cfr sys-export` stalling on large clusters

### LINKED ISSUE #429: CFR: `ctk cfr sys-export` stalls after 24% ...

florinutz:

`ctk cfr sys-export` reads `sys.jobs_log` with an unbounded `SELECT *`. Every node keeps its own log, capped at 5% of its heap, so the node answering has to hold all of them in one HTTP response, and the parent circuit breaker refuses it:

```
CircuitBreakingException[Allocating 239mb for 'parent: http-result' failed, breaker would use 1gb in total. Limit is 972.7mb ...]
```

The breaker counts real heap use, garbage included, so an idle node goes on refusing everything, down to the small error it owes the client. That is the state behind [#429](https://github.com/crate/cratedb-toolkit/issues/429#issuecomment-5679274922): the node goes quiet and the export hangs.

### The change

The export no longer asks for the whole log. It collects the 1000 most recent entries of `sys.jobs_log` and `sys.operations_log`, `--log-limit` sets another number, and `manifest.json` records it.

### How the entries are collected

@bgunebakan, [who finally caught the so-far mysterious failure](https://github.com/crate/cratedb-toolkit/issues/429#issuecomment-5679274922), suggests `ORDER BY ended DESC LIMIT 1000`. That bounds what the client receives, but not what the cluster has to handle. No node knows whether its own entries are among the newest 1000 across the cluster, so every node sends its own newest 1000: twenty nodes move 20,000 rows of statement text for the 1000 that are kept.

Now the export asks twice instead. The first query is the same sort, over times alone:

```sql
SELECT ended FROM sys.jobs_log ORDER BY ended DESC LIMIT 1 OFFSET 999
```

The nodes still merge 20,000 entries, but a time is one number where a statement runs to tens of kilobytes. What comes back is a single value: when the 1000th newest entry ended. The second query reads the rows from that moment on, which each node applies to its own log:

```sql
SELECT * FROM sys.jobs_log WHERE ended >= 1789544981690
```

Those two queries fetch one page of 1000 entries. A higher `--log-limit` runs them again for the next page, over the entries older than the window just read:

```sql
SELECT ended FROM sys.jobs_log WHERE ended < 1789544981690 ORDER BY ended DESC LIMIT 1 OFFSET 999
SELECT * FROM sys.jobs_log WHERE ended < 1789544981690 AND ended >= 1789544980928
```

The loop ends at the limit, or earlier when the log runs out, so no single response carries more than a page.

A read also waits 120 seconds at most, since a node that stops answering otherwise blocks the export with nothing to show (`?timeout=` in the cluster URL still wins). A log table that fails mid-way keeps the pages that arrived, listed in the manifest under `data_partial` with the error.

### Evidence

The reproduction from #429 — one node, 1 GB heap, `stats.breaker.log.jobs.limit=400mb`, 390 MB of statements in the log — runs through:

```
Successfully exported 38 tables from sys, information_schema (18 with data, 0 partial, 1 skipped, 0 schema and 0 data failures, 0 definition failures)
```

I did not tackle the idea of collecting only the shards that are not started. If support needs it, it should come later, separate from the fix of this bug.

## Checklist

 - [x] Link to issue this PR refers to (if applicable): Closes #429


bgunebakan:
@hammerhead Could you please have a look, Is it works as you expected?

hammerhead:
I tested it on a ten-node cluster where each node had its jobs log at 10.000 entries, and exporting worked fine. I now get 1002 rows total, across the different nodes. With `--log-limit 200` it was limited to 205 rows total.

hammerhead (approved):

bgunebakan (approved):
Thank you for fixing this! works good

PR #906, opened 2026-09-16
Testing: Build the CrateDB testcontainer on the upstream Testcontainers module

### LINKED ISSUE #905: Testing: Use the upstream Testcontainers CrateDB module
Testcontainers for Python ships a [CrateDB](https://github.com/testcontainers/testcontainers-python/commit/0976c7e7ec3455e90731c84d4ab65de378d350d7) module since 4.15.0. 

Toolkit still carries its own CrateDB container class, and its `testing` extra caps `testcontainers<4.15`, so the projects that take their test container from here cannot use the upstream module: `sqlalchemy-cratedb`, `commons-codec`, `crash`, `cratedb-dlm` and  `cratedb-examples`.

Goals:
- everyone gets the upstream container
- `cratedb-toolkit` becomes based on on the new upstream and also adds what upstream lacks: keeping a container alive between runs, skipping tests when Docker is not running, the nightly default image, `reset()` and some database helpers. 
- `CrateDBTestAdapter` and the `cratedb_service` fixture should keep working as they do today, so the repos above need no changes of their own.

Done when the `testing` extra allows testcontainers 4.15 and a release is out. #883 bumps the same pin. After the release, `cratedb-dlm` can drop its `testcontainers<4.15`.

florinutz:
Closes #905.

`CrateDBContainer` now extends the CrateDB module of Testcontainers for Python instead of carrying its own container class, and the `testing` extra moves from `testcontainers<4.15` to `testcontainers[cratedb]>=4.15,<4.16` (this supersedes #883). 

The extra is needed because the upstream module reaches CrateDB over HTTP through `httpx`.

### What changed

Startup, readiness and URL building come from upstream. Keep-alive, skipping when Docker is down, the nightly default image, the two extra server options, the host port mapping, `reset()`, `.database` and the `cratedb_service` fixture stay as they were. The signatures called by our downstream repositories also stay unchanged.

### Once this is released, 
- crate/cratedb-examples#2059 can use toolkit's `CrateDBTestAdapter` in its examples
- `cratedb-dlm` can drop `testcontainers<4.15`.

the change in `util.py` comes because 4.15's type stubs reject the `volumes` argument (the error that blocked #883)


cratedb_toolkit/testing/testcontainers/cratedb.py
```diff
@@ -56,20 +54,21 @@ class CrateDBContainer(DockerSkippingContainer, KeepaliveContainer, DbContainer)
...
 
     def __init__(
         self,
         image: str = "crate/crate:nightly",
-        ports: Optional[dict] = None,
-        user: Optional[str] = None,
-        password: Optional[str] = None,
-        dbname: Optional[str] = None,
-        cmd_opts: Optional[dict] = None,
+        ports: t.Optional[dict] = None,
```
bgunebakan: I'd keep import statement on top as `from typing import Optional` because keeping all imports as `Optional` is better and align with whole codebase.
```suggestion
        ports: Optional[dict] = None,
```

cratedb_toolkit/testing/testcontainers/cratedb.py
```diff
@@ -14,20 +14,18 @@
...
 import logging
 import os
-import re
-from typing import Optional
+import typing as t
 
-from testcontainers.core.generic import DbContainer
-from testcontainers.core.wait_strategies import LogMessageWaitStrategy
-from testcontainers.core.waiting_utils import wait_for_logs
+from testcontainers.community.cratedb import CrateDBContainer as UpstreamCrateDBContainer
```
bgunebakan, line 20: Why don't you directly use`CrateDBContainer`?

```suggestion
from testcontainers.community.cratedb import CrateDBContainer
```

tests/retention/test_cli.py
```diff
@@ -43,7 +50,7 @@ def test_setup_brief(caplog, cratedb, settings):
     assert result.exit_code == 0
 
     assert cratedb.database.table_exists(settings.policy_table.fullname) is True
-    assert 1 <= len(caplog.records) <= 2
+    assert toolkit_records(caplog) == []
```
bgunebakan: `caplog.set_level(logging.ERROR, "sqlalchemy")` sets the capture handler's level globally so `caplog.records` is always empty, this tests won't fail. maybe we need to improve or change this test.

hammerhead (approved):
Interesting, I wasn't aware that the toolkit's test components are used in downstream projects. Maybe worth documenting one day, feels rather hidden at the moment 🙂.

bgunebakan (changes requested):
Thank you, I've added some comments.

cratedb_toolkit/testing/testcontainers/cratedb.py
```diff
@@ -149,16 +151,8 @@ def get_connection_url(self, dialect: str = "crate", host: Optional[str] = None)
...
-        wait_for_logs(self, predicate=self._wait_strategy, timeout=15)
-
-    def _configure_wait_condition(self):
-        """Wait for CrateDB node to be fully started."""
-        # TODO: Better use a network connectivity health check?
-        #       In `testcontainers-java`, there is the `HttpWaitStrategy`.
-        # TODO: Provide a client instance.
-        self.waiting_for(LogMessageWaitStrategy(re.compile(r"o.e.n.Node.*started")))
+        # The HTTP interface may be configured to a port other than the upstream default.
+        HttpWaitStrategy(self.port_to_expose).for_status_code(200).wait_until_ready(self)
```
bgunebakan: this increases the timeout silently from 15s to 120s, that could increase the wait times.

cratedb_toolkit/testing/testcontainers/cratedb.py
```diff
@@ -56,20 +54,21 @@ class CrateDBContainer(DockerSkippingContainer, KeepaliveContainer, DbContainer)
     CRATEDB_PASSWORD = os.environ.get("CRATEDB_PASSWORD", "crate")
     CRATEDB_DB = os.environ.get("CRATEDB_DB", "doc")
     KEEPALIVE = asbool(os.environ.get("CRATEDB_KEEPALIVE", os.environ.get("TC_KEEPALIVE", False)))
-    CMD_OPTS = {
-        "discovery.type": "single-node",
-        "node.attr.storage": "hot",
-        "path.repo": "/tmp/snapshots",
-    }
+    # `testcontainers` builds its command line from pairs; `cmd_opts` merges over these, by key.
```
bgunebakan, line 61: PR body says _"The signatures called by our downstream repositories also stay unchanged."_ that's true for signatures but this class attribute has changed. I think needs to mention in changelog.

cratedb_toolkit/testing/testcontainers/cratedb.py
```diff
@@ -140,7 +142,7 @@ def get_connection_url(self, dialect: str = "crate", host: Optional[str] = None)
         """
         # TODO: When using `db_name=self.CRATEDB_DB`:
         #       Connection.__init__() got an unexpected keyword argument 'database'
-        return super()._create_connection_url(
+        return self._create_connection_url(
```
bgunebakan: Why did you change to referring the parent class here?

cratedb_toolkit/testing/testcontainers/cratedb.py
```diff
@@ -101,12 +102,12 @@ def __init__(
         self.port_to_expose, _ = list(self.port_mapping.items())[0]
```
bgunebakan, line 106: The upstream container also introduces another argument `port` and after this PR we will have 2 arguments that store port, and two of them can be different. `port_to_expose` exist because there wasn't `port` arg before. but still [util.py:50](https://github.com/crate/cratedb-toolkit/pull/906/changes#diff-6a0018d03bd53f9ceaae22614f1b7a634497c2c784d7dcfd2f7ad26d4f7c9015R50) fall back to `port_to_expose`. so if someone try different port, it will always use hardcoded port value from upstream.

bgunebakan (changes requested):

cratedb_toolkit/testing/testcontainers/cratedb.py
florinutz: will do, but the repo runs both styles

cratedb_toolkit/testing/testcontainers/cratedb.py
florinutz: our subclass is also called `CrateDBContainer`

tests/retention/test_cli.py
florinutz: you're right. I have now set `caplog.at_level(logging.DEBUG)` for these tests

cratedb_toolkit/testing/testcontainers/cratedb.py
florinutz: yes, this 120 was testcontainer's default (`testcontainers_config.timeout`). Made it explicit now: `CrateDBContainer.STARTUP_TIMEOUT = 60`, set on the wait strategy. It will increase the wait time indeed, but only in the case of a long workload or error (e.g. failing start), right?

the check polls and returns as soon as the HTTP interface answers

cratedb_toolkit/testing/testcontainers/cratedb.py
florinutz: you're right, added, thanks!

cratedb_toolkit/testing/testcontainers/cratedb.py
florinutz: leftover. reverted, thanks.

cratedb_toolkit/testing/testcontainers/cratedb.py
florinutz: Good catch, fixed with `self.port = self.port_to_expose`

florinutz:
CI fails because of main

bgunebakan (approved):
Thanks, looks good

